PR #16, opened 2022-08-02
Add errors

chandr-andr:


pyproject.toml
```diff
@@ -12,6 +12,8 @@ license = "LICENSE"
 python = "^3.7"
 typing-extensions = ">=3.10.0.0"
 pydantic = "^1.6.2"
+redis = "^4.3.4"
```
s3rius, line 15: Эт зачем это тут?

chandr-andr: А как ошибки редиски ловить?

s3rius, line 15: Please delete this. 

We don't want any extra dependency in core repo.

pyproject.toml
```diff
@@ -40,6 +42,10 @@ implicit_reexport = true
 allow_untyped_decorators = true
 warn_return_any = false
 
+[[tool.mypy.overrides]]
```
s3rius, line 45: image

taskiq/abc/broker.py
```diff
@@ -121,13 +123,18 @@ async def kiq(
...
         :returns: taskiq task.
         """
         logger.debug(
             f"Kicking {self.task_name} with args={args} and kwargs={kwargs}.",
         )
         message = self._prepare_message(*args, **kwargs)
-        await self.broker.kick(message)
+        try:
+            await self.broker.kick(message)
+        except AMQPConnectionError as exc:
```
s3rius, line 136: Please catch `Exception` not specific exceptions. 

What if someone will implement kafka broker? We don't want to have kafka in dependencies in taskiq and moreover we don't want to add extra except clause.

s3rius (changes requested):
You have to be more generic. Since taskiq gives people ability to easily add support for their technologies. We don't want to catch specific exceptions. We want to catch all of them.

Also please remove redundant dependencies. Core repo must be as small as possible.

pyproject.toml
chandr-andr: Deleted

pyproject.toml
chandr-andr: Deleted it 1 minute ago))

taskiq/abc/broker.py
chandr-andr: Changed specific exceptions to `Exception`

taskiq/cli/async_task_runner.py
```diff
@@ -277,3 +280,4 @@ async def async_listen_messages(  # noqa: C901, WPS210, WPS213
             await broker.result_backend.set_result(message.task_id, result)
         except Exception as exc:
             logger.exception(exc)
+            raise ResultSetError() from exc
```
s3rius, line 283: I think it would be better not to raise exception here. Since it will kill the worker's process.

s3rius (changes requested):
A little change to make workers more stable.

taskiq/cli/async_task_runner.py
chandr-andr: Okay, but maybe we can change logger message to more understandable? I think that people need to explicitly understand that the broker is broken and cannot accept the message

chandr-andr:
> A little change to make workers more stable.

Done, please have a look

taskiq/cli/async_task_runner.py
```diff
@@ -275,4 +275,9 @@ async def async_listen_messages(  # noqa: C901, WPS210, WPS213
         try:
             await broker.result_backend.set_result(message.task_id, result)
         except Exception as exc:
-            logger.exception(exc)
+            logger.exception(
```
s3rius, line 278: You better do this:

```python
logger.error(
    "Can't set result in %s result backend: \n %s.",
    broker.result_backend,
    exc,
    exc_info=True
)
```

s3rius (approved):
LGTM.

PR #35, opened 2022-08-23
Updated broker API.

s3rius:
Signed-off-by: Pavel Kirilin <win10@list.ru>


s3rius:
Now brokers don't have to yield new message, they have to call callback. It allows us to make message processing fully asynchronous.

PR #58, opened 2023-01-04
docs: rebuild docs

Mister-Hope:
This pr includes:

- yarn -> pnpm, pnpm has better performance on both install and startup
- deps update
- typo fixes
- broken link fix
- replace `@vuepress/plugin-search` with `vuepress-plugin-search-pro` which can index all content


s3rius:
Thank you very much! Awesome!

Mister-Hope:
Emm, just have another check, I missed this line:

https://github.com/taskiq-python/taskiq/blob/10bd19dc9df92083ee4b31749cabffc553a607c4/.github/workflows/release.yml#L39

You may need to fix it to `docs/.vuepress/dist` I guess, or you should add `dist: './docs-dist'` in `.vuepress/config.ts`, sorry for that.

s3rius:
I've fixed it. Don't worry.

PR #64, opened 2023-03-19
Refaсtored process manager. 

### LINKED ISSUE #63: Worker hangs sometimes
Sometimes we fall into state, when taskiq worker hangs and doesn't react to signals like SIGTERM.

We need to fix it, since it's really annoying. Maybe we should differently work with child processes.

s3rius:
Refactored process management for workers. 

Now we use ublocking event queue and a big fat custom event loop.

It closes #63. Worker hanged, because observer thread was blocking on event queue until all events are processed. To solve this problem, I had to rewrite the whole process management logic from scratch.

### DIFF
--- a/taskiq/cli/worker/run.py
+++ b/taskiq/cli/worker/run.py
@@ -1,5 +1,4 @@
 import asyncio
-import enum
 import logging
 import signal
 import sys
@@ -19,14 +18,6 @@ try:
     import uvloop  # noqa: WPS433
 except ImportError:
     uvloop = None  # type: ignore
-
-
-class ProcessAction(enum.Enum):
-    """Enumberate of possible events."""
-
-    FULL_RELOAD = enum.auto()
-    SHUTDOWN = enum.auto()
-    RELOAD_ONE = enum.auto()
 
 
 logger = logging.getLogger("taskiq.worker")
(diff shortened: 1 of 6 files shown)


chandr-andr (approved):
Very good, but not the best (rewrite it in rust plz).

taskiq/cli/worker/run.py
```diff
@@ -225,43 +261,42 @@ def run_worker(args: WorkerArgs) -> None:  # noqa: WPS213
...
-            name=f"worker-{process}",
-        )
-        work_proc.start()
-        logger.debug(
-            "Started process worker-%d with pid %s ",
-            process,
-            work_proc.pid,
-        )
-        worker_processes.append(work_proc)
+    action_queue: "Queue[Tuple[ProcessAction, Optional[int]]]" = Queue(-1)
```
booqoffsky, line 274: What about creating a class with event args? I think this will make the code easier to read.

For example :
```
class ReloadOneActionArgs:
    worker_num: int
```
or some common class for all events

s3rius: Maybe you're right. I'll refactor it a little more.

s3rius: Please check it. I created a separate class for managing processes and classes for all actions. Now it looks a bit better.

booqoffsky:
A W E S O M E

booqoffsky (approved):

PR #81, opened 2023-03-28
Use a separate thread to watch for dead processes

Kolaer:
Should help with #63 

"Works on my machine" (c), but I have a somewhat unusual setup and I don't have enough knowledge of the project to properly test this (I've run a bunch of manual tests, but I'm still not convinced that this works in all configurations).

@s3rius can you look into this and test on your setup as well? We tried a similar approach together and I remember that it didn't work for some reason.


s3rius:
Sure. Let me check it.

s3rius (changes requested):
The code looks good to me. But it doesn't work.

PR #84, opened 2023-04-01
Updated InMemoryBroker

s3rius:
This PR makes inmemory broker more robust, because now we don't await for tasks to be executed, we execute them in a true async way.


booqoffsky (approved):

booqoffsky:
LGTM

PR #90, opened 2023-04-03
Separation of the receiver from the CLI

hugiron:


s3rius (changes requested):
Since you've moved the `async_listen_messages` function inside of the receiver, you need to delete original `async_listen_messages` function. Because we don't want to have deadcode in our codebase, right?

hugiron:
@s3rius already

s3rius (approved):
Ah, I see. My bad. Yeah, it's good.

s3rius:
But change brach to develop.

PR #93, opened 2023-04-05
asyncio.wait replaced with Semaphore, fixed compatibility issues

### LINKED ISSUE #91: Use semaphore to replace the current
https://github.com/taskiq-python/taskiq/blob/318b25b40e2cc9753c6e4c76b462a9627ecb0b62/taskiq/cli/worker/async_task_runner.py#L47-L49
We can use `asyncio.Semaphore` to replace the current scheme, and it is not recommended to set `max_async_tasks` less than 1, the infinite number of tasks will only cause unexpected situations.
Use `asyncio.Semaphore` to limit the number of tasks, assign task scheduling to the event loop, and avoid using `asyncio.wait` to increase power consumption concurrently.

vvanglro:
1. use semaphore, add test
2. change default max_async_tasks num
3. fix py39 has removesuffix
4. fix pre-commit

Please review it.
This PR will close #91.

### DIFF
--- a/taskiq/abc/broker.py
+++ b/taskiq/abc/broker.py
@@ -30,7 +30,7 @@ from taskiq.formatters.json_formatter import JSONFormatter
 from taskiq.message import BrokerMessage
 from taskiq.result_backends.dummy import DummyResultBackend
 from taskiq.state import TaskiqState
-from taskiq.utils import maybe_awaitable
+from taskiq.utils import maybe_awaitable, remove_suffix
 
 if TYPE_CHECKING:  # pragma: no cover
     from taskiq.abc.formatter import TaskiqFormatter
@@ -57,7 +57,7 @@ def default_id_generator() -> str:
     return uuid4().hex
 
 
-class AsyncBroker(ABC):  # noqa: WPS230
+class AsyncBroker(ABC):
     """
     Async broker.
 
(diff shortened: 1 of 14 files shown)


s3rius:
@vvanglro, can you please rebase onto develop? I updated CI so jobs must run now.

vvanglro:
CI looks like it needs your approval.

s3rius:
You've done a great job. But isn't 10 tasks is too low number by default? And also, now you cannot set maximum number of tasks to `-1`.

vvanglro:
There are 10 concurrent tasks by default for each process, and 50 for 5 processes, which I think is quite a lot. Of course, users can also pass in the number by themselves.

s3rius:
Yes, you're right. It looks like it would work well. Thank you.

s3rius (approved):
Looks good to me.

taskiq/abc/broker.py
```diff
@@ -234,15 +234,22 @@ def inner(
...
+                            )
+                        else:
+                            fmodule = ".".join(
+                                sys.argv[0]
+                                .removesuffix(
+                                    ".py",
+                                )
+                                .split(
+                                    os.path.sep,
+                                ),
```
booqoffsky, line 251: Hello! What do you think about not splitting the logic into two parts and using only a method that is suitable for all versions of python?

vvanglro: Yes, it's great 😌

booqoffsky (approved):
A W E S O M E

PR #115, opened 2023-05-05
fix: execution time retains 2 digits

vvanglro:
image

When using the memory broker, because the task is executed quickly, the execution time becomes scientific notation, so I think it is enough to keep 2 decimal places


s3rius (approved):
Fair point. LGTM.

PR #119, opened 2023-05-06
fix: kiq and kicker type hint

vvanglro:
This pr fixes the problem that kiq and kicker do not have type hints in pycharm.

image
image

### DIFF
--- a/taskiq/decor.py
+++ b/taskiq/decor.py
@@ -72,7 +72,7 @@ class AsyncTaskiqDecoratedTask(Generic[_FuncParams, _ReturnType]):
         self: "AsyncTaskiqDecoratedTask[_FuncParams, _ReturnType]",
         *args: _FuncParams.args,
         **kwargs: _FuncParams.kwargs,
-    ) -> AsyncTaskiqTask[_ReturnType]:
+    ) -> AsyncTaskiqTask[Any]:
         ...
 
     async def kiq(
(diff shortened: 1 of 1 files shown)


s3rius:
Doesn't it work with plain self?

vvanglro:
Yes, due to the use of generics, the previous type prompt cannot be correctly recognized.

vvanglro:
At least in the pycharm I use, there is no prompt.

s3rius:
It's seems like a bug of pycharm. 

Removing generics from DecoratedTask will break autosuggestions for other editors. For example, every editor that uses PyLance for auto suggestions can even show you arguments of your task for kiq method. 

```python
@broker.task
async def mytask(a: int, b: int):
    """Some docs"""


async def main():
    await broker.startup()

    mytask.kiq()

```

image

image


If you remove generics that won't be possible. They already has an issue for that: [link](https://youtrack.jetbrains.com/issue/PY-57765/Code-completion-doesnt-work-for-class-functions-decorated-with-PEP612-annotated-decorators).

s3rius:
I would suggest you to try VSCode. Their autosuggestions work better than in pycharm for new PEPs.

vvanglro:
emm, it may be a problem with pycharm.
We can look into it again, anyway, our goal is to let users have the correct type hints when using it.
I can't stay up any longer, I'm going to bed first.🥱

s3rius:
Good night.

vvanglro:
😅 I tried again. Before pycharm repairs support, we can make a patch-like method to do type hints. Can you see if this will affect the previous hints?

PR #121, opened 2023-05-08
feat: set/get progress of task execution

Sobes76rus:
Hello there! I would like to add some methods to set/get progress of tasks executions
In Celery u can do it with `update_state`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1537,14 +1537,14 @@ pbr = ">=2.0.0,<2.1.0 || >2.1.0"
 
 [[package]]
 name = "taskiq-dependencies"
-version = "1.2.2"
+version = "1.2.3"
 description = "FastAPI like dependency injection implementation"
 category = "main"
 optional = false
 python-versions = ">=3.7,<4.0"
 files = [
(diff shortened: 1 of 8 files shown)


s3rius:
Hi and thanks for creating this PR. I have a question regarding API for setting a progress from within a task. How's it going to be implemented?

s3rius:
I guess, we can introduce new dependency for base context, that can work with progresses.

Sobes76rus:
I can use `context.broker.result_backend.set_progress`
Or provide some addition function to dependencies for it

s3rius:
@Sobes76rus, I guess it's better to have a class-based dependency with tab-completed methods.

```python
from taskiq import Context

class TaskState(enum.Enum):
	PENDING = enum.auto()
	STARTED = enum.auto()
	...


class ProgressTracker:
	def __init__(self, context: Context = Depends()):
		self.context = context
	
	def set_progress(self,  state: TaskState, meta: Dict[str, Any]) -> None:
		...

```

Sobes76rus:
Ok, ill do it

PR #123, opened 2023-05-09
feat: custom reciever

Sobes76rus:
Hello there! I would like to let ppl use custom receiver if they need it.


s3rius:
Okay. It looks good. What do you think of adding a way to provide it with custom arguments `--receiver_arg "arg1=my_value"` ? It may be implemented in another PR or issue.

s3rius (approved):

vvanglro:
Well, it’s better to be an abstract class, specify some methods that need to be implemented, and users can inherit the implementation.

PR #126, opened 2023-05-10
fix: retry_middleware

Sobes76rus:
`retry_middleware` is not working as expected
maybe we should ignore `set_result` in case of retry?


s3rius:
Let's not ignore it, but set it to PROGRESS and in meta you could put current retry and total retries.

s3rius:
But in this case, we restrict users from using retryable functions and custom progresses. Because it may lead to undefined behavior.

Sobes76rus:
Yeah, Is'nt ignore `set_result` will be better?
If use progress state we still should rewrite `is_ready` function to check status

Sobes76rus:
It'is not backward compatible

s3rius:
That's true. Yes, we can make it after we integrate the progress feature. We just need to keep it in mind.

Sobes76rus:
Then check my PR in taskiq-dependencies pls, i need it to implement progress feature)
fix https://github.com/taskiq-python/taskiq-dependencies/pull/11

s3rius:
@Sobes76rus I checked it already. I just want to make it work even better. By the way, how are you gonna use it in context of dependencies? I didn't think about generics.

Sobes76rus:
I can define type of progress for example

Sobes76rus:
We can do it in another way
Like in Celery we can raise `RetryError` exception and keep it in `TaskiqResult`

Then we can check if tasks is completed in `is_ready` if task's result was set with this exception
Or, check it in `wait_result` with some arg like `check_retry=True` which i think will be better

In that case we dont require set/get progress feature
So, in that way i suggest to improve `TaskiqResult` with new fileds like `exc` and `traceback`

s3rius:
We can put a traceback in result, because it's easy to serialize, but I'm not sure about exceptions. Custom exception can contain unserializable data. Which might be a problem.

Sobes76rus:
There are `find_pickleable_exception` method in Celery
> With an exception instance, iterate over its super classes (by MRO) and find the first super exception that’s pickleable

I believe they serialize only pickable exceptions

s3rius:
That makes sense, actually. This way we can even propagate exceptions to client-side.

Sobes76rus:
[Celery exceptions serialization](https://docs.celeryq.dev/en/stable/internals/reference/celery.utils.serialization.html#celery.utils.serialization.UnpickleableExceptionWrapper)

s3rius (approved):

PR #128, opened 2023-05-11
feat: non-blocking tasks

Sobes76rus:
non-blocking sleep through a temporary increase in the number of running tasks on demand of the user.
here is some use cases
1. some task A is waiting for result of another task B. User can request a temporary increase in the number of tasks, so this task A does not block the execution of task B because of `max_async_tasks`. Example in tests with `max_async_tasks = 1`
2. i have a real case scenario where very long running tasks sleeps for a very long time and there is a need to keep their scope, so i increase the max number of running tasks while they asleep

### DIFF
--- a/taskiq/context.py
+++ b/taskiq/context.py
@@ -1,12 +1,22 @@
-from contextlib import _AsyncGeneratorContextManager
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Optional
+
+from typing_extensions import TypeAlias
 
 from taskiq.abc.broker import AsyncBroker
 from taskiq.exceptions import NoResultError, TaskRejectedError
 from taskiq.message import TaskiqMessage
 
 if TYPE_CHECKING:  # pragma: no cover
+    from contextlib import _AsyncGeneratorContextManager
+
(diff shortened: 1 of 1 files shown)


lembregtse:
@s3rius are there plans to have this merged / or implemented differently? We have a use-case as well that triggers this that we noticed since we have now implemented the max_async_tasks option. 

We will "workarround" by separating them into two different workers for now.

s3rius:
There are plans to implement it differently. There were some attempts. 

The main idea of another approach is to not give just a sleep function, but an ability to decrease counter of running tasks by the time of executing a function. 

I've created an issue for that problem here: #258.

Sobes76rus:
Hello there @s3rius ! I changed approach to dealing with idle tasks
Now u can do smth like this
```[python]
async def task(idle: TaskIdler = Depends()):
    async with idle():
        awat smth()
```
Pls take a look

Sobes76rus:
Hello, @s3rius. Pls take a look at the new implementation

Sobes76rus:
@s3rius A friendly reminder to check this PR :D

Sobes76rus:
Hello there @s3rius! Pls give it a try =)

PR #130, opened 2023-05-12
feat: set/get progress

Sobes76rus:
properly created PR instead of https://github.com/taskiq-python/taskiq/pull/121


Sobes76rus:
@s3rius Ok, i did it =)

taskiq/depends/progress_tracker.py
```diff
@@ -0,0 +1,99 @@
...
+        Set progress of task execution.
+
+        :param state: TaskState value.
+        :param meta: progress of task.
+        """
+        ...
+
+    async def set_progress(
+        self,
+        state: Union[TaskProgress[_ProgressType], TaskState, str],
```
s3rius, line 79: I guess, leaving only TaskState or str is sufficent for any usecase.

s3rius: Personally I would not allow str values, because it adds ambiguity. Since, if you check results you want to know what to check. Adding string states adds complexity for every check. 

But on the other hand, I see why people may want it.

s3rius: @taskiq-python/core-team, what're your thoughts on this?

taskiq/depends/progress_tracker.py
```diff
@@ -0,0 +1,99 @@
...
+        meta: Union[_ProgressType, Undefined] = undefined,
```
s3rius, line 80: Why do we need undefined value? Why not just use None? Because we won't set anything for built-it states and if people want  to set something, they can set a value.

taskiq/abc/result_backend.py
```diff
@@ -50,3 +60,25 @@ async def get_result(
         :param with_logs: if True it will download task's logs.
         :return: task's return value.
         """
+
+    async def set_progress(
+        self,
+        task_id: str,
+        progress: "TaskProgress[_ProgressType]",
```
s3rius, line 67: Just set generic parameter to Any, because ResultBackend doesn't need to know about exact type. And we cannot parametrize it with this type, since it can vary from task to task.

```suggestion
        progress: "TaskProgress[Any]",
```

taskiq/depends/progress_tracker.py
```diff
@@ -0,0 +1,99 @@
...
+_ProgressType = TypeVar("_ProgressType")
+
+
+class Undefined(enum.Enum):
+    """For undefined value."""
+
+    undefined = ...
+
+
+undefined = Undefined.undefined
```
s3rius, line 21: Why not just set it to ellipsis? 

```suggestion
undefined = ...
```

s3rius (changes requested):
Hi and thanks for your time and crating this PR. I have a questions related to generics. How can you use the _ProgressType? Because on client side there's no information about progress type that this task can possibly have. 

You can add a parameter to the task decorator, to parameterize `AsyncTaskiqDecoratedTask` with progress type. Because now this generic adds no information for users about progress type of a task.

taskiq/depends/progress_tracker.py
Sobes76rus:
```
@broker.task
await def task(progress: ProgressTracker[int] = Depends()): 
    await progress.set_progress(123)
```

taskiq/depends/progress_tracker.py
Sobes76rus: In case user define Optional type of progress

taskiq/abc/result_backend.py
Sobes76rus: Will be easy to switch to PEP696

taskiq/depends/progress_tracker.py
Sobes76rus: Cannot use Ellipsis in Union type

taskiq/depends/progress_tracker.py
Sobes76rus: It is just safe guard

s3rius: In that case, they can define undefined value by themself. Maybe just use `...`.

taskiq/abc/result_backend.py
Sobes76rus: ResultBackend does't know about it right now, but then PEP696 is implemented it will know

Sobes76rus: I wanna add second generic parameter in ResultBackend 
But for background compatibility it should have default value

taskiq/depends/progress_tracker.py
Sobes76rus: Ok, ill change it

Sobes76rus:
For example

```@broker.task
await def task(progress: ProgressTracker[int] = Depends()): 
    await progress.set_progress('progress', 123)
```

taskiq/depends/progress_tracker.py
s3rius: Okay. I guess I overcomplicate things.  We can just check if it's equals to some value.

```python
task = await my_task.kiq()
# Here we cannot say what type parameter does progress have.
progress = task.get_progress()
if progress.state == "progress":
	print("Current progress", 123)

```

taskiq/abc/result_backend.py
s3rius: We don't need to switch to this pep for result backend. My point is that we don't need to set it as generic value, because it's not used in any place. This generic doesn't have any value to neither developer nor users.

Sobes76rus: Ok

taskiq/depends/progress_tracker.py
s3rius: Why? 

```python
from typing import TYPE_CHECKING, Union

if TYPE_CHECKING:
    ELLIPSIS = ellipsis


a: "Union[str, ELLIPSIS]" = "meme"
```

`ellipsis` builtin only available during type_checking. You can use it to remember it somewhere for later use. The only thing here is that this type hint should be a string, because ELLIPSIS is not available at runtime .

taskiq/depends/progress_tracker.py
Sobes76rus: Done

taskiq/abc/result_backend.py
Sobes76rus: Done

taskiq/depends/progress_tracker.py
Sobes76rus: Done

Sobes76rus:
@s3rius Hi! Sorry to be annoying, but could you give me some feedback please. =)

s3rius (commented):
Sorry, I really had no time lately. 

The request looks good. But I have some questions, regarding types. 

You introduced a `TaskProgress` which is generic over `_ProgressType`. So generic types are used by IDEs to help developers get information about types they're working with or to help programs cast some random structure to some specific type. The problem with `_ProgressType` is that it's not used anywhere. I guess you imagined usage like this:

```python
from pydantic import BaseModel
from myproj.tkq import broker

class MyProgress(BaseModel):
	id: int
	name: str

async def my_task(ptrack: ProgressTracker[MyProgress] = Depends()):
	for i in range(100):
		ptrack.set_progress("heavy_calcs", MyProgress(id=i, name=f"name_{i}"))
		# Do some heavy work.
```

And it looks good and easy. But types work only inside of the task. Check out this code snippet:

```python
	task = await my_task.kiq()
	progress = task.get_progress()
	progress.meta.<TAB> # Here you won't get any auto completion.
```

To solve this issue, you have to add new generic to `AsyncTaskiqTask` task and a way to set this generic. The easiest way is to add a parameter to `task` method of the `AsyncBroker` interface and add one overload that make use of this generic. A small example of how types should look like so people would get suggestions based on their types:

```python
class AsyncTaskiqDecoratedTask(Generic[_FuncParams, _ReturnType, _ProgressType]):
    @overload
    async def kiq(
        self: "AsyncTaskiqDecoratedTask[_FuncParams, Coroutine[Any, Any, _T], _ProgressType]",
        *args: _FuncParams.args,
        **kwargs: _FuncParams.kwargs,
    ) -> AsyncTaskiqTask[_T, _ProgressType]:
        ...

class AsyncTaskiqTask(Generic[_Task[_ReturnType], _ProgressType]):
    async def get_progress(self) -> "Optional[TaskProgress[_ProgressType]]":
        """
        Get task progress.
        :return: task's progress.
        """
        return await self.result_backend.get_progress(self.task_id)

class AsyncBroker(ABC):
    @overload
    def task(
        self,
        task_name: Callable[_FuncParams, _ReturnType],
		progress_type: Optional[Type[_ProgressType]] = None,
        **labels: Any,
    ) -> AsyncTaskiqDecoratedTask[_FuncParams, _ReturnType, _ProgressType]:  # pragma: no cover
        ...

```

Sobes76rus:
@s3rius Hello there! I have resolved conflicts with latest version and make some changes. Could u pls take another look =)

kice:
~~I was looking for a progress report feature to show a progress bar, and found this. This feature name should change to "Task state reporting"~~

My very one-sided opinion on this feature is: pulling states from worker instead of pushing to result backend.

In case of way too many tasks, I think state reporting will easily over overwhelmed result backend. This might be solved by using reporting to a broker instead of a result backend, or have a limiter built-in.

For me, I might report what percentage the task is finished, which I should report as often as possible; and if there is not that much states, it is not necessary to report.

Also, I think state reporting needs to be one-way communication, or it will confuse people: should I use "state reporting" to handle error and retry, or use a middleware.

~~Maybe the end goal is to make a worker monitoring for taskiq.~~

Sobes76rus:
@s3rius hello there! Could you please take another look at this request

taskiq/depends/progress_tracker.py
```diff
@@ -0,0 +1,72 @@
...
+        meta: Optional[_ProgressType] = None,
+    ) -> None:
+        """Set progress.
+
+        :param state: TaskState or str
+        :param meta: progress data
+        """
+        if meta is None:
+            progress = await self.get_progress()
+            meta = progress.meta if progress else meta
```
s3rius, line 56:
```suggestion
            meta = progress.meta if progress else None
```

It's just a bit more obvious.

s3rius (changes requested):
I really like the idea from here. And the implementation is actually good. But here's some things which I try to accomplish and currently had no luck. 

I tried to make some mechanism to add type suggestions to progresses. Because i generally dislike working with raw dicts instead of python types. I will try to do more research on that later. 

As for the PR, I think it's great. The only one thing which is missing for me is setting the progress for tasks inside the receiver. When task was started, when task was executed etc.

Also, you can show usage of retry by updating simple-retry middleware.

taskiq/brokers/inmemory_broker.py
```diff
@@ -79,6 +81,37 @@ async def get_result(
         """
         return self.results[task_id]
 
+    async def set_progress(
+        self,
+        task_id: str,
+        progress: TaskProgress[Any],
+    ) -> None:
+        """
+        Set progress of task exection.
```
kyboi, line 90:
```suggestion
        Set progress of task execution.
```

kyboi:
Is the idea of this PR just to know the overall progress of a task as started/failure/success/retry or can an actual status message be passed from the worker task and fetched from the backend/broker (e.g. progress percentage or stage to then be displayed to user)?

s3rius:
@kyboi, yes. Mainly for progress.

.github/workflows/test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         py_version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-        pydantic_ver: ["<2", ">=2,<3"]
+        pydantic_ver: ["<2", ">=2.5,<3"]
```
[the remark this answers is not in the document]
Sobes76rus: Fix Optional field with validate_default only performing one field validation
by @sydney-runkle in https://github.com/pydantic/pydantic-core/pull/1002

taskiq/abc/result_backend.py
```diff
@@ -50,3 +54,25 @@ async def get_result(
         :param with_logs: if True it will download task's logs.
         :return: task's return value.
         """
+
+    async def set_progress(
```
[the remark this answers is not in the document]
Sobes76rus: Then `set_result` should have a similar behaviour i believe

Sobes76rus:
@s3rius 
I think it should be up to the user to decide how to use the progress tracker. 
The overall status should be implemented with some `ProgressMiddleware`, not in the receiver. Or there should be a way to disable it.
In other cases, the user can pass custom progress messages using the `ProgressTracker` dependency.

Sobes76rus:
@kyboi 
my intension was to use custom status messages.
overall status should be optional middleware

taskiq/abc/result_backend.py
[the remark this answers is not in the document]
s3rius: Actually yes, but I might fix it for all abstracts in another PR.

s3rius:
@Sobes76rus, I guess the progress middleware can be implemented in subsequent request, so it's fine to not set statuses here.

s3rius (approved):
Looks good to me. Will merge a bit later.

PR #134, opened 2023-05-15
Fix typos and command in docs

tschiolborg:
While looking at the docs, I found that the commands 
```
taskiq test_project.broker:broker test_projec.submodule.tasks test_projec.utils.tasks
```
```
taskiq test_project.broker:broker -fsd
```

were missing the `worker` argument.

While fixing it, I also fixed some obvious typos.


s3rius:
Thank you very much. That's really helpful.

s3rius (approved):

PR #145, opened 2023-05-26
feat: add package version

vvanglro:
Add package version.


s3rius:
Thanks for your contribution. I have a question. Why do we need this variable, since you can easily get it by yourself in more unified way using importlib.metadata?

vvanglro:
Because I think this is the standard for third-party libraries.

And people are used to this method to check the version.

image

s3rius (approved):
Okay, makes sense. Let's merge it. 

Thank you for your contribution.

PR #154, opened 2023-06-12
fix: requeue message

vvanglro:
It seems that there is no need to copy here, If you need to regenerate a message object here you should use deepcopy.


s3rius (approved):
Hi, @vvanglro, you're right. I was thinking about deepcopy, but then decided to modify it in-place. 

Thanks for spotting it.

Issue #158, opened 2023-06-28
Logging configuration should not impact loggers outside `taskiq`


MuriloScarpaSitonio:
My team uses Datadog (DD) for monitoring and we deploy using containers with default configuration - which means that everything emitted to `stdout`/`stderr` is pushed to DD as log.

The worker cli entrypoint (`taskiq.cli.worker.run.run_worker`) currently defines logging configuration using `logging.basicConfig` and I think this should be avoided because doing so configures the root logger (`logging.root`), impacting all other loggers in the hierarchy.

In my use case, I have a logger with a custom formatter to emit logs in a format thad DD can parse. As `taskiq` is using `logging.basicConfig` this adds a handler to the root logger then all logs emitted by my logger are also being emitted by the root handler (`StreamHandler`). This is causing duplicates in DD and, on top of that, the logs emitted by the root handler are being marked with `ERROR` regardless the actual log level because this handler uses `sys.stderr` as a stream.

My suggestion: get rid of `logging.basicConfig` and define a root log for `taskiq` instead of applying rules to root logger of the whole application - i.e. do something like this instead:

```python
logger = logging.getLogger("taskiq")
handler = logging.StreamHandler(stream=sys.stdout)
handler.setFormatter(logging.Formatter(args.log_collector_format))
logger.setLevel(logging.getLevelName(args.log_level))
``` 

For the record, my current solution was to not propagate logs emitted by my custom handler...
```python
logger = logging.getLogger("my_cutom_logger")
logger.propagate = False
...
``` 
... and regarding the `sys.stderr` thing, I'm currently running the worker with `--log-level=${TASKIQ_LOG_LEVEL:-ERROR}` so no big deal (although I wouldn't know how to easily adjust this if I was using a lower value for `--log-level`).

s3rius:
Hi, and thanks for raising this issue! Looks really reasonable, and as I know, many people use super custom log formats. 

I guess it's a good idea to get rid of this. Will try to do something with it. Or, alternatively you can try submitting a PR that works in exactly your case.

s3rius:
@MuriloScarpaSitonio, please take a look at #159. If it works for you, we're going to release new version right after your approve.

MuriloScarpaSitonio:
Thanks, @s3rius! I just left a comment there. Unfortunately, I don't have much time this week otherwise I could try to push a PR. Next week will be more relaxed for me, so if it's ok for you I think I can work on a PR.

PR #159, opened 2023-06-28
Added optional logging configuration. 

s3rius:
This request adds parameter that disables automatic logging configuration for taskiq workers. 

This PR should solve #158.

### DIFF
--- a/taskiq/cli/worker/run.py
+++ b/taskiq/cli/worker/run.py
@@ -153,10 +153,13 @@ def run_worker(args: WorkerArgs) -> None:  # noqa: WPS213
 
     :raises ValueError: if reload flag is used, but dependencies are not installed.
     """
-    logging.basicConfig(
-        level=logging.getLevelName(args.log_level),
-        format="[%(asctime)s][%(name)s][%(levelname)-7s][%(processName)s] %(message)s",
-    )
+    if args.configure_logging:
+        logging.basicConfig(
+            level=logging.getLevelName(args.log_level),
+            format="[%(asctime)s][%(name)s][%(levelname)-7s]"
+            + "[%(processName)s] %(message)s",
+        )
+    logging.getLogger("taskiq").setLevel(level=logging.getLevelName(args.log_level))
     logging.getLogger("watchdog.observers.inotify_buffer").setLevel(level=logging.INFO)
     logger.info("Starting %s worker processes.", args.workers)
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -226,13 +226,13 @@ files = [
 
 [[package]]
 name = "click"
-version = "8.1.3"
+version = "8.1.4"
 description = "Composable command line interface toolkit"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "click-8.1.3-py3-none-any.whl", hash = "sha256:bb4d8133cb15a609f44e8213d9b391b0809795062913b383c62be0ee95b1db48"},
-    {file = "click-8.1.3.tar.gz", hash = "sha256:7682dc8afb30297001674575ea00d1814d808d6a36af415a82bd481d37ba7b8e"},
+    {file = "click-8.1.4-py3-none-any.whl", hash = "sha256:2739815aaa5d2c986a88f1e9230c55e17f0caad3d958a5e13ad0797c166db9e3"},
+    {file = "click-8.1.4.tar.gz", hash = "sha256:b97d0c74955da062a7d4ef92fadb583806a585b2ea81958a81bd72726cbb8e37"},
 ]
 
 [package.dependencies]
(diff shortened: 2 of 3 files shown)


taskiq/cli/worker/run.py
```diff
@@ -153,10 +153,12 @@ def run_worker(args: WorkerArgs) -> None:  # noqa: WPS213
 
     :raises ValueError: if reload flag is used, but dependencies are not installed.
     """
-    logging.basicConfig(
-        level=logging.getLevelName(args.log_level),
-        format="[%(asctime)s][%(name)s][%(levelname)-7s][%(processName)s] %(message)s",
-    )
+    if args.configure_logging:
+        logging.basicConfig(
```
MuriloScarpaSitonio, line 157: I think this detaches the logs from the root logger but then I can't control what level `taskiq` should emit, right?
I could achieve that by doing something like...
```python
logging.getLogger("taskiq").setLevel(<level>)
```
...in my application.

But this is super counter-intuitive, specially because we'd still have `--log-level` as an argument.

From my perspective, we should still be able to control the log level from the cli and `taskiq` should simply detaches its loggers from root.

s3rius: I implemented this [here](https://github.com/taskiq-python/taskiq/compare/0f1fda2d87d5580450708312bf65534f60db8c32..ca5d3fe0711ed3033760ed62c1b7361eb6c09a2b). 

Still, I don't get the idea of detaching logger from others. Imagine you have loguru configured. In can intercept log calls, but still you would have to manually add taskiq in list of intercepted loggers, because it's not propogated to the root logger. I'm not quite sure if it's okay.

Is it a problem for your case, if we do not detach `taskiq` named logger from the root one?

MuriloScarpaSitonio: For my application is not a concern because I decided to set `propagate=False` in my handlers.

I've never used `loguru` so for me it's hard to say. But from your description I think it makes more sense to explicitly add an "interception" than explicitly disabling propagation from my logger on my application.

`Taskiq` is just an implementation detail (even being the most important dependency - don't get me wrong :)) for my application so I think it makes more sense that my application should "say" if `taskiq` logs should appear than having `taskiq` impacting my logs. I have the same train of thought for a new dependency: I wouldn't like to have to remember that I have to disable its loggers but rather I'd like to be able to easily enable if I want it.

On top of that, by configuring the root logger you also "force" that loggers from other dependencies to also show up, because not all of them explicitly set `propagate` to `False` - meaning that the stream is prone to "pollution" since loggers from all other packages will start to be propagated to the `root` logger. This (potentially, I'm not entirely sure) also means that I lose granularity over the levels of those other loggers: I can't easily set the level of `taskiq` logs to `INFO` and `ddtrace` to `ERROR`, for example, because by setting `taskiq` logs to `INFO` it actually means `root` logs to `INFO` and because `ddtrace_logger.propagate = True` I think I'll also receive logs when `ddtrace_logger.level = INFO`. 

And, to be honest, I've never seen a dependency configure the root logger (let's take a look at uvicorn [here](https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L74C1-L74C15) and [here](https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L379), for instance).

Does all of this make sense?

s3rius: Yes it makes sense. But the main idea is that taskiq has a solid and ready-to-use configuration that would work for any project and you can change every aspect of it with few configuration parameters. It only configures logger for its own processes, it doesn't affect client-size, because it run this code only on worker processes. It configures root logger, because in many cases people don't configure logging. This is an out of the box simple configuration that you can use easily. If you want to configure logging by yourself, you can turn off automatic configuration and do whatever you want. 

If it doesn't make sense to you, please make a request with your solution that works. We would continue discussion there. Thank you for finding time for a discussion.

MuriloScarpaSitonio: Yes, being able to completely turn on and off logging configuration also makes sense. 

I would only disagree with "it doesn't affect client-size": although I acknowledge that this is true for most applications, it's not the case for many others - i.e. those larger systems whose services are decoupled and communication are done trough messaging. For those systems, the client side is the worker processes as the application entrypoint is that task that will process the message published in the related broker.

s3rius: Okay. But even in this case when worker processes are clients. How would it affect them if there would be no logging configuration? If you have an Idea of how you'd like this feature to be implemented, please provide me with simple example. Because I like the Idea and don't see edge cases in current implementation. If you do, please tell me what you see and maybe how to fix it.

MuriloScarpaSitonio: Oh it was just a comment. I think this solution and your train of thought is great.

s3rius: Thank you very much for the discussion. It was really enjoyable. I will merge this pull request. But still, you can raise issues any time.

PR #162, opened 2023-07-05
fix: worker hangs sometimes

Sobes76rus:
issues: https://github.com/taskiq-python/taskiq/issues/63 https://github.com/taskiq-python/taskiq/pull/81

child process inherits signal handlers from parent

the problem is that the `terminate` signal was triggered before the signal handler of the child process was set, we must set it immediately after the process starts

and I suggest telling the parent process that the worker is ready before the parent continues with the event queue.
or at least wait a while for the worker process to set the signal handler
or we can handle it in parent signal handler, and raise an error if it is not a main process

### DIFF
--- a/taskiq/cli/worker/process_manager.py
+++ b/taskiq/cli/worker/process_manager.py
@@ -1,7 +1,7 @@
 import logging
 import signal
 from dataclasses import dataclass
-from multiprocessing import Process, Queue
+from multiprocessing import Process, Queue, current_process
 from time import sleep
 from typing import Any, Callable, List, Optional
 
@@ -82,6 +82,7 @@ class ReloadOneAction(ProcessActionBase):
         new_process.start()
         logger.info(f"Process {new_process.name} restarted with pid {new_process.pid}")
         workers[self.worker_num] = new_process
+        sleep(0.1)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


taskiq/cli/worker/process_manager.py
```diff
@@ -82,6 +82,7 @@ def handle(
         new_process.start()
         logger.info(f"Process {new_process.name} restarted with pid {new_process.pid}")
         workers[self.worker_num] = new_process
+        sleep(0.1)
```
s3rius, line 85: Why?

s3rius (changes requested):
Actually good suggestions. Can you implement these? 

I would implement it using multiprocessing Queue. Since we know how many workers we would like to have, so we can just wait for n messages from this queue, where n is the number of worker processes.

taskiq/cli/worker/process_manager.py
s3rius: It's better to use multiprocessing events then.

Sobes76rus: I did it right now

Sobes76rus:
OK, I did it through `multiprocess.Event'.

s3rius (approved):
I think it's great. Thank you.

PR #164, opened 2023-07-10
Bug/task-scheduled-by-time-runs-multiple-times

### LINKED ISSUE #163: Schedule task from another task
Hi, there! First of all, thank you for `taskiq`. I found it really useful and ease to work with. 

And I wonder, is it possible to schedule task from another executable task. For example:

```python
@broker.task('do_something')
async def do_something(arg: int):
    print('do_something...', arg)

@broker.task('tasks_smart_scheduler', schedule=[{"time": datetime.utcnow()}])
async def tasks_smart_scheduler():
    print('tasks_smart_scheduler...')

    # Create do_something task with scheduler label to run it later.
    later = datetime.utcnow() + timedelta(seconds=10)
    decorated_task = broker.task(
        f'do_something:{later}',   # pass custom name to avoid conflicts with initial `do_something` task
        schedule=[
            {
                "time": datetime.utcnow(),
                "args": 123
            }
        ],
    )(do_something)

    assert decorated_task in broker.available_tasks.values()  # Task added to broker successfully
```

- `decorated_task` is not enqueued here, I'm expecting that scheduler will do that on proper time (later). If I wont to enqueue it's wright now, I could write following. But it's not what I'm trying to do.

```python
await decorated_task.kiq(123)
```

So, I run both `worker CLI` and `scheduler CLI`. But, obviously, it does not work, because them running in **separate** processes, and `scheduler` is knowing **nothing** about new scheduled task just created at worker process.

Is it possible to communicate between scheduler and worker? Run them simultaneously in one process somehow? Or it's an anti-pattern at all?

MishaVyb:
## Bug

Task with `time` (not `cron`) option should be enqueued only once, but `run_scheduler` sending it to broker every loop iteration (every minute).

## TDD

I'm starting from describing a test case and it falls down. As expected. For writing a proper test, I had to add 2 new dev dependency. I hope you found them applicable for Taskiq, otherwise this test could take up to 2-3 minutes and it's not conveniently.

## Implementation

The solution is quite simple (writing a test was more complicated, as usual 😅). I move responsibility for sending tasks when it becomes **ready** from `run_scheduler` main function to `scheduler` itself. 

```python
if ready:
    logger.info("Sending task %s.", task.task_name)
    loop.create_task(scheduler.on_ready(task))
```

I think, it makes sense and it's allowed me to define `pre_send` and `post_send` abstract methods. And then I implement `post_send` for `LabelScheduleSource`. It's just remove trigger `time` from schedule label list. That's all. 

At the beginning I was thinking about adding extra field like `"was_send": True / False`. But it seemed more complicated and may cause side effects. My logic is simple, when task have triggered by time, it's not scheduler task anymore, so we remove that particular trigger.

## Unexpected consequences
The only case this solution won't work as expected, it's when `while` loop in `run_scheduler` will iterate again **earlier** 
than `schedules_updater` updates tasks list. And task that we just have sent will be sent again. 

For default options it wan't happen, as we update tasks every 30.0 seconds, while `run_scheduler` iterates every minute. But what if somebody define `refresh_delay` with value more then 1 minute 🤔... I was not debugging that case, but theoretically it breaks my implementation.  

Closes #163


taskiq/scheduler/scheduler.py
```diff
@@ -56,3 +59,17 @@ async def startup(self) -> None:  # pragma: no cover
...
         await self.broker.startup()
+
+    async def on_ready(self, task: ScheduledTask) -> None:
+        """
+        This method is called on task is ready to be enqueued.
+
+        It's triggered on proper time depending on `task.cron` or `task.time` attribute.
+        :param task: task to send
+        """
+        await task.source.pre_send(task)
```
s3rius, line 70: Can you wrap it in `maybe_awaitable`? This will add you posibility to add sync functions.

```python
	if task.source.pre_send != TaskiqScheduler.pre_send:
		await maybe_awaitable(task.source.pre_send(task))
```

taskiq/scheduler/scheduler.py
```diff
@@ -56,3 +59,17 @@ async def startup(self) -> None:  # pragma: no cover
...
+        await AsyncKicker(task.task_name, self.broker, task.labels).kiq(
+            *task.args,
+            **task.kwargs,
+        )
+        await task.source.post_send(task)
```
s3rius, line 75: Do the same for post-send.

taskiq/schedule_sources/label_based.py
```diff
@@ -1,9 +1,12 @@
+from logging import getLogger
 from typing import List
 
 from taskiq.abc.broker import AsyncBroker
 from taskiq.abc.schedule_source import ScheduleSource
 from taskiq.scheduler.scheduler import ScheduledTask
 
+logger = getLogger(__name__)
```
s3rius, line 8: What name is it going to be? If it starts with taskiq (it probably should), then it's ok.

taskiq/cli/scheduler/args.py
```diff
@@ -3,13 +3,14 @@
...
 from taskiq.cli.common_args import LogLevel
+from taskiq.scheduler.scheduler import TaskiqScheduler
 
 
 @dataclass
 class SchedulerArgs:
     """Arguments for scheduler."""
 
-    scheduler: str
+    scheduler: str | TaskiqScheduler
```
s3rius, line 13: Do not use 3.10+ notation. Since taskiq should work even with older versions of python.

s3rius:
You can run `pre-commit -a` locally to check all lint problems.

MishaVyb:
yes, I see, but I'm not sure it runs tests against all python versions. Anyway, the lust commit should pass all tests

MishaVyb:
Hmm, it looks really weird... All tests are passed locally, but it's failed on CI. And fail reason is `asyncio.TimeoutError`, but it is exactly that behaviour we are expecting by wrapping it in context

```python
with pytest.raises(TimeoutError):
    await asyncio.wait_for(event.wait(), 2.0)
```

Let's leave it until tomorrow, I'll look again, what's wrong. Maybe I should reimplement this test. Also could we run workflow fully (for now all except 3.9 were canceled)? It helps me to see, will it failed on other python versions or not.

s3rius:
Hi, @MishaVyb. I checked the PR once again. Replace `TimeoutError` with `asyncio.TimeoutError`.

Also, I've added `tox` support to run test against different python version by one command. It's not yet available in docs, but you can follow this [instruction](https://github.com/taskiq-python/taskiq/blob/develop/docs/contrib.md) to setup everything for testing all python versions locally. 

Also, you have some problems with typing. Python 3.8 doesn't support `dict` and `list` types. You have to specify `List` and `Dict` explicitly.

taskiq/schedule_sources/label_based.py
s3rius: I checked it. The name is ok.

MishaVyb:
@s3rius, sorry for wasting your time, and thanks for bringing `tox` - useful tool! I didn't know about that. And thank you for your suggestion, it works. So, I think, PR is finally done and could be merged 😌

At least all tests passed by `tox` locally and I hope, it will passed here, on github actions. Also add `pre-commit run --all-files` at `tox.ini` to be sure, all linter tests are passed on all python versions.

s3rius (approved):
Thanks for your PR, @MishaVyb. That's totally not a waste of time, since it makes the lib better.

MishaVyb:
Great! If be honest, it was my fist contribution for open source ever, so It's kind of excited moment for me. Also, I've just found another small issue in log output. Maybe, I'll come back with another PR later.

Issue #192, opened 2023-08-17
Run workers and schedulers programmatically


igoose1:
Worker and scheduler had been written to be cli-first. I suggest to move all logic to Worker/Scheduler classes and make [cli module](https://github.com/taskiq-python/taskiq/tree/master/taskiq/cli) just an interface to them.

Currently, to launch taskiq workers I should provide `WorkerArgs` to `run_worker`. It's a huge class without default values because it's hard-coded to be used with `argparse`.

### Example
I want to show how [uvicorn](https://www.uvicorn.org/) (ASGI web server implementation) works. Server can be launched from a command line:

```sh
$ uvicorn main:app
```

There's a code equivalent:

```python
import uvicorn

async def app(scope, receive, send):
    ...

if __name__ == "__main__":
    uvicorn.run("main:app", port=5000, log_level="info")
```

s3rius:
It makes sense. Will try to add it later.

dtatarkin:
I faced the same problem and solved it by extracting peace of code from Taskiq to separate function

```
from concurrent.futures import ThreadPoolExecutor

from taskiq import AsyncBroker
from taskiq.cli.worker.run import shutdown_broker
from taskiq.receiver import Receiver


async def run_worker(broker: AsyncBroker, worker_params: TaskiqWorkerParams):
    try:
        with ThreadPoolExecutor(max_workers=worker_params.max_sync_workers) as pool:
            receiver = Receiver(
                broker=broker,
                executor=pool,
                validate_params=True,
                max_async_tasks=worker_params.max_async_tasks,
                max_prefetch=worker_params.max_prefetch,
            )
            broker.is_worker_process = True
            await receiver.listen()
    finally:
        await shutdown_broker(broker, worker_params.shutdown_timeout)
```

s3rius:
Now we have a nice API for dynamic environments and now you can run brokers and schedulers right inside your application. 

However it was not possible to implement exactly the same functionality. If you want to run exactly the same functionality as in `taskiq worker`. You can do something like that.

```python
from taskiq_redis import ListQueueBroker

from taskiq.cli.worker.args import WorkerArgs
from taskiq.cli.worker.run import run_worker

broker = ListQueueBroker("redis://localhost")


def main():
    run_worker(WorkerArgs(broker="b:broker", modules=[], fs_discover=True))


if __name__ == "__main__":
    main()
``` 

But it's exactly the same and will completely block the main function.

PR #202, opened 2023-09-24
Add the ability to dynamically create multiple brokers with different sets of registered tasks.

dtatarkin:
I'm using multiple dynamically created brokers with different sets of registered tasks. This is a "killer feature" of Taskiq for me.
But it seems that although the Taskiq architecture is able to implement such a use case, there is a micro bug that hinders.
Each time I register a task for an instance of the AsyncBroker class, this task is registered as available for all brokers.
This one line change will open a very powerful feature of dynamically created brokers with different set of registered TaskiqDecoratedTask tasks.

### DIFF
--- a/taskiq/abc/broker.py
+++ b/taskiq/abc/broker.py
@@ -68,13 +68,13 @@ class AsyncBroker(ABC):
     in async mode.
     """
 
-    available_tasks: Dict[str, AsyncTaskiqDecoratedTask[Any, Any]] = {}
-
     def __init__(
         self,
         result_backend: "Optional[AsyncResultBackend[_T]]" = None,
         task_id_generator: Optional[Callable[[], str]] = None,
     ) -> None:
+        self.available_tasks: Dict[str, AsyncTaskiqDecoratedTask[Any, Any]] = {}
+
         if result_backend is None:
             result_backend = DummyResultBackend()
         else:


s3rius:
Hello! And thanks for your PullRequest. That's awesome that you proving and improving flexibility of the taskiq library.

s3rius:
Actually, I'm quite curious how exactly you use dynamic brokers. That would be really nice of yours if you could provide me with some example. A gist, maybe. Thanks in advance.

dtatarkin:
I'm using Taskiq as an inter-service API in a micro-service architecture. I had a positive experience with Celery, but now want to migrate to Taskiq as a more flexible, lightweight and async solution. The idea is that a service registers tasks as API endpoints and other services can call these tasks.
In such an approach, no REST, gRPC, etc. protocol implementation is required, since Taskiq + Pydantic handle (de)serialization and RPC calls perfectly. Service discovery is also not required since Redis/RabbitMQ handles this part. With Taskiq it is also possible to implement both "task queue" (ListQueueBroker) and "publisher-subscriber" (PubSubBroker) patterns, which I really miss in Celery.

This is how I call Task from another service without accessing that service's source code.
```
    async def get_assignments(
        self,
        request: GetAssignmentsRequest,
        timeout: float | None = None,
    ) -> TaskiqResult[GetAssignmentsResponse]:
        kicker = AsyncKicker(
            task_name='backoffice:get_assignments',
            broker=self._broker,
            labels={},
        )
        task = await kicker.kiq(request=request)
        result = await task.wait_result(timeout=timeout or self._params.timeout)
        return result
```

dtatarkin:
Some of my micro-services are created dynamically as async task in the same python thread. Therefore, I faced a problem when Taskiq tasks are registered at AsyncBroker class level rather than instance level. My use case also involves class method functions as Taskiq tasks, which is more susceptible to the problem I am trying to solve. BTW, this issue is closely related to the functionality mentioned in this request https://github.com/taskiq-python/taskiq/issues/192.

s3rius:
That's really interesting. Thank you for sharing! But actually global tasks were made for this specific feature, so tasks are interchangeable between brokers. Here's why. 

The taskiq has a cool feature that allows you to define tasks that doesn't have broker by default and cannot be kicked without a specific broker. 

You can do it using [async_shared_broker](https://taskiq-python.github.io/available-components/brokers.html#async-shared-broker-and-shared-tasks). This is not a real broker but can be used to define global tasks which can be kicked and executed by other brokers.  

For example, you have your function defined with `async_shared_broker`. 

```python
from taskiq import async_shared_broker


@async_shared_broker.task
async def get_assigments(id: int) -> None: ...
```

Now we can use this task to actually send to other brokers. 

```python
get_assigments.kicker().with_broker(my_broker).kiq(id=1)
```

Now the mypy will validate this call and arguments and this approach will be less error-prone. Also, we might listen to this task for specific broker. And if any broker will receive this task it will make a lookup in it's available tasks.

But here goes the possible pitfall if we're going to register tasks only for specific brokers. If you defined task with implementation the broker won't be able to execute it. As an example of global task registry usage: We have an ability to pipeline tasks one after another by using `taskiq-pipelines` library. And because we want to execute pipelines without being attached to a specific broker we need to have ability to register tasks that were not registered directly using `broker.task`. Here's an [example of such task](https://github.com/taskiq-python/taskiq-pipelines/blob/4c3f0aa2b1f1a6f270b0112fe47e67aa5490cc33/taskiq_pipelines/steps/mapper.py#L22).

I'm not sure how to solve this issue right now. Maybe create separate global task registry that is used only by `async_shared_broker` and all tasks that were registered directly will be registered in a specific broker. We need to test this solution.

s3rius:
@dtatarkin, can I ask you to write me directly [on telegram](https://t.me/s3rius_san) or on our [discord server](https://discord.gg/GT7npe9cJq)? 

After the discussion I will summarize it here or in another discussion.

dtatarkin:
Closed in favor of a more correct approach https://github.com/taskiq-python/taskiq/pull/203

PR #203, opened 2023-09-25
Added local task registry, improved shared_tasks API. 

s3rius:
As discussed with @dtatarkin in #202 we need to have a difference between tasks registered by user and shared, that were created using `async_shared_broker`.


dtatarkin:
Thank you very much, this solution works well for me!

PR #206, opened 2023-09-26
Make  `dynamic-brokers.md` easier to read

Kolaer:
I think after these changes, the article is somewhat easier to read. There are still some rough edges, but we can focus on them later.

I can go through the docs and fix everything that I can find, but it'll take some time, so let's start with some smaller changes first.


Kolaer:
Hi,  @s3rius

Please proof-read the changes. I'm not 100% sure that they are in sync with the code/examples.

s3rius (approved):
It's easier to read, indeed. Thank you very much.

PR #214, opened 2023-10-10
Implement customizable serializer

asvetlov:
Pre-requirement for https://github.com/taskiq-python/taskiq-pipelines/issues/13 

The main idea is: 
1. `JSONFormatter` exists for the backward compatibility.
2. `ProxyFormatter` is used by default, which operates over `broker.serializer` API
3. The default serializer is `JSONSerializer` which is a thin wrapper around plain `json.dumps()` and `json.loads()`
4. `TaskiqSerializer` base class works with `bytes`, not `str`. The broker message type is a bytes anyway, plus formats like msgpack and cbor work with bytes anyway.

I can implement `cbor` and `msgpack` serializers as well if you are ok with optional dependencies (say, `cbor2` is required only if a user explicitly imports `taskiq.serializers.cbor_serializer:CBORSerializer`.


s3rius (approved):
Now I see the point crystally clear. These changes will create a simpler API for defining your own serializers. Thank you very much. 

I'm totally ok with optional dependencies. In the next iteration I was planning to add a lot of optional depndencies to make installation process easier.

s3rius:
To fix `mypy` please consider running `poetry update`. 

To fix failing tests, make them async and mark with `pytest.mark.anyio`.  This problem happens because initialize semaphore in InmemoryBroker. This works in py3.10 and above, but doesn't work for older versions. We will fix it.

asvetlov:
tests should be green now, at least locally I don't see any problems

s3rius:
Since these mypy problems are totally minor, we can merge it now. I'll fix it later.

s3rius:
Thank you very much for your contribution.

asvetlov:
You are welcome! The next step is PR for taskiq-pipelines

asvetlov:
Could you please consider publishing a new release?

It is a pre-requirement for conversion taskiq-pipelines to use new serializers.

s3rius:
Sure.

s3rius:
Changes available in https://github.com/taskiq-python/taskiq/releases/tag/0.9.3.

PR #222, opened 2023-10-17
Fixed when messages are acked.

s3rius:
This PR fixes the acknowledgements behavior. If worker goes down when executing tasks, message won't be acknowledged untill it executed.

### DIFF
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -131,14 +131,14 @@ class Receiver:
             taskiq_msg.task_id,
         )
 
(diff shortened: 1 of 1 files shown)


s3rius:
Sure. Can you please clarify what issues have you ran into?

thebalaa:
In our [Matrix powered TaskIQ broker](https://github.com/fractalnetworksco/taskiq-matrix) a slow task is being run multiple times because we expected the worker to ack task before running them.

I was under the impression that if you were concerned with the outcome of a task you could await the result from the result backend. Does it make sense?

PR #232, opened 2023-10-20
Added pickle serializer.

s3rius:
Pickle can serializer any types and might be useful for systems where security is not a concern.

### DIFF
--- a/taskiq/serializers/__init__.py
+++ b/taskiq/serializers/__init__.py
@@ -3,10 +3,12 @@ from .cbor_serializer import CBORSerializer
 from .json_serializer import JSONSerializer
 from .msgpack_serializer import MSGPackSerializer
 from .orjson_serializer import ORJSONSerializer
+from .pickle import PickleSerializer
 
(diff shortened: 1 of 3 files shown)


aryadovoy:
Hi @s3rius,
Could you show how to use this serializer with `RedisAsyncResultBackend`? I tried like this:
```python
broker.with_serializer(serializer=PickleSerializer())
```
But got error: `_picke.UnpicklingError: invalid key, '{'.` Do I need something else?

s3rius:
@aryadovoy, the redis-result-backend uses different serializer, so you need to pass it directly as the parameter of your resultBackend. https://github.com/taskiq-python/taskiq-redis/blob/e30ed080f4df584f8de03535bffd3a0c1c090586/taskiq_redis/redis_backend.py#L58

aryadovoy:
@s3rius, looks like `PickleSerializer` is default for `RedisAsyncResultBackend`, so no changes needed and can pass any objects as result in Redis?

PR #234, opened 2023-10-24
Scheduled task is a pydantic model now.

s3rius:
This is made because of serialization for datetimes. It makes it easier to serialize and deserialize. Now Schedule sources have more freedom in serialization.


taskiq/scheduler/scheduled_task.py
```diff
@@ -1,11 +1,12 @@
 import uuid
-from dataclasses import dataclass, field
+from dataclasses import field
```
vvanglro, line 2: Here you should use pydantic's Field, In the ScheduledTask class need to be replaced with Field.

s3rius: True.

s3rius: Also, I've missed support for different pydantic versions. Now if should be fine.

PR #237, opened 2023-11-02
Changed deprecated logger.warn to logger.warning

DeanNeaht:


s3rius:
Thanks for your contribution. Is there only one place where warn is used?

DeanNeaht:
yes

PR #243, opened 2023-11-26
Added non-string labels in tasks.

s3rius:
This problem was highlighted by @asvetlov in the https://github.com/taskiq-python/taskiq-pipelines/pull/14.

I was thinking about possible APIs for implementation this functionality and came up with this one. Now, along with dict of labels we store dict of labels types. To minimize amount of space for this information I've chosen to use ints over anything else. This solution is not perfect, but it's the most elegant solution I could come up with. 

Possible improvements for the future:
* Add ability to add custom label types to not parse them manually.
* Remove `parse_labels` method from `TaskiqMessage` and move it somewhere else (possibly in formatters)

This request should be backward compatible, but might brake some custom middlewares.


s3rius:
@asvetlov, mind taking a look at the request? I'd be very grateful, since this functionality affects your PR as well. 

Thanks in advance.

Issue #244, opened 2023-11-28
Possible scheduled tasks duplication


stinovlas:
As far as I know, current design of scheduler works like this:
1. send all tasks that are scheduled to current minute
2. wait until next minute start (while asynchronously keeping an eye on scheduled tasks)
3. repeat send with delays with one minute interval

This creates a possible duplication of tasks. If I restart scheduler in a minute where there is a cron scheduled task, this task is sent twice (once before the restart and once after). On the other hand, it allows the scheduler to be restarted without missing a task, assuming the restart takes less then one minute to perform (which is quite reasonable assumption for scheduler).

Ideally, tasks should be idempotent, but that's not always possible. It would be great if taskiq kept in mind whether particular task has already been sent to worker queue in one minute and if so, wouldn't send it again after restart. This is obviously not possible with `LabelScheduleSource` since it doesn't have any permanent storage, but it might be possible with `taskiq_redis.RedisScheduleSource`.

What do you think about this? Should I propose an enhancement in `taskiq_redis`? Does something need to be done in `taskiq` itself?

s3rius:
I totally agree that for some systems that might be the case. And here's what I think. We have middlewares and `post_send` events on each scheduler. You can create a simple storage that might be attached to all schedule sources.  

```python
import datetime
from typing import Any, Coroutine, List
from redis.asyncio import Redis
from taskiq import ScheduleSource
from taskiq.scheduler.scheduled_task import ScheduledTask
from taskiq.utils import maybe_awaitable


class ScheduleProtector(ScheduleSource):
    def __init__(self, redis_url: str, source: ScheduleSource) -> None:
        self.redis = Redis.from_url(redis_url)
        self.source = source

    def get_schedules(self) -> Coroutine[Any, Any, List[ScheduledTask]]:
        return self.source.get_schedules()

    def add_schedule(self, schedule: ScheduledTask) -> Coroutine[Any, Any, None]:
        return self.source.add_schedule(schedule)

    def delete_schedule(self, schedule_id: str) -> Coroutine[Any, Any, None]:
        return self.source.delete_schedule(schedule_id)

    async def pre_send(self, task: ScheduledTask) -> Coroutine[Any, Any, None]:
        res = await self.redis.get(task.schedule_id)
        if res is not None:
            try:
                ts = int(res)
                if ts > int(datetime.datetime.utcnow().timestamp()):
                    raise RuntimeError("The task has already been sent.")
            except ValueError:
                pass
[...]

PR #254, opened 2023-12-05
Add scheduler pre_send exception handling

stinovlas:
Followup to #244.

This PR implements a new exception `ScheduledTaskCancelledError` which can be raised in `ScheduleSource.pre_send` method.

This exception is then caught and processed in `TaskiqScheduler.on_ready`. This way, schedule source can decide upon invocation whether the task should be actually performed or not.


s3rius:
Looks good to me.

PR #257, opened 2023-12-06
Update task patterns

### LINKED ISSUE #256: Tasks pattern includes .tox and other directories
I wanted to use task autodetection, but it failed because the worker tried to load packages from `.tox` in my working directory:

```
taskiq worker --fs-discover --tasks-pattern 'mypackage/**/*.py' mypackage.tasks:broker
```

This actually fails with uncaught exception, because `.tox.py311.lib.python3.11.site-packages.mypackage.tasks` is not a valid module path.

The problems seems to be that `taskiq.cli.utils.import_tasks` uses `rglob` to generate files list. There is no way for me to exclude directories I don't want to be traversed. Using `Path().glob(pattern)` instead of `Path().rglob(pattern)` looks like a reasonable fix – that way everyone can choose whether they want to prefix their pattern with `**/` or not.

This would be a breaking change. It makes sense, but is breaking nonethless. I would personally be for implementing it, but the call is @s3rius's, not mine. If we are changing it anyway, I'd also suggest the possibility to add multiple task patterns (by repeating the option) and search all of them.

If you agree with the proposal, I'll handle the implementation.

stinovlas:
Closes #256.

As per #256, i changed the `rglob` pattern search to `glob` and added a possibility to specify `--tasks-pattern` option multiple times.

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -26,7 +26,7 @@ That's why taskiq can auto-discover tasks in current directory recursively.
 We have two options for this:
 
 - `--tasks-pattern` or `-tp`.
-  It's a name of files to import. By default is searches for all `tasks.py` files.
+  It's a glob pattern of files to import. By default it is `**/tasks.py` which searches for all `tasks.py` files.
 - `--fs-discover` or `-fsd`. This option enables search of task files in current directory recursively, using the given pattern.
 
 ### Acknowledgements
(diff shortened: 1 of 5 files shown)


s3rius:
Actually I like the approach, because it will make it task searching more flexible. I will update minor version because of this PR, Since it might break apps that were using custom  `--tasks-pattern`.

s3rius:
Also, can you make `import_tasks` to accept either sequence or string as pattern? Because I was using this function in some taskiq-related third-party libs.

s3rius:
Thank you very much for your contributions.

Issue #261, opened 2023-12-11
Using Sentry SDK in worker leads to crash


itssimon:
The Taskiq workers crash immediately if I initialise the Sentry SDK in the worker startup handler.

```
$ taskiq worker --ack-type when_executed --reload apitally.tasks:broker                                                                4s  cloud  base 20:37:47
[2023-12-11 20:40:48,284][taskiq.worker][INFO   ][MainProcess] Starting 2 worker processes.
[2023-12-11 20:40:48,284][root][WARNING][MainProcess] Reload on change enabled. Number of worker processes set to 1.
[2023-12-11 20:40:48,290][taskiq.process-manager][INFO   ][MainProcess] Started process worker-0 with pid 30331 
objc[30331]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[30331]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
[2023-12-11 20:40:50,301][taskiq.process-manager][INFO   ][MainProcess] worker-0 is dead. Scheduling reload.
[2023-12-11 20:40:51,314][taskiq.process-manager][INFO   ][MainProcess] Process worker-0 restarted with pid 30338
objc[30338]: +[NSNumber initialize] may have been in progress in another thread when fork() was called.
objc[30338]: +[NSNumber initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
[2023-12-11 20:40:53,330][taskiq.process-manager][INFO   ][MainProcess] worker-0 is dead. Scheduling reload.
[2023-12-11 20:40:54,338][taskiq.process-manager][INFO   ][MainProcess] Process worker-0 restarted with pid 30342
objc[30342]: +[NSNumber initialize] may have been in progress in another thread when fork() was called.
objc[30342]: +[NSNumber initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

The worker startup handler looks like this:

```python
@broker.on_event(TaskiqEvents.WORKER_STARTUP)
async def on_worker_startup(state: TaskiqState) -> None:
    sentry_sdk.init(
[...]

PR #262, opened 2023-12-11
Changed multiprocessing method on macos.

s3rius:
Changed a MacOS multiprocessing start method to spawn, because of the issue #261.


s3rius:
The presented solution haven't solved the issue. Closing.

Issue #229, opened 2023-10-19
taskiq.receiver.receiver.Receiver.prefetcher attached to a different loop


enlike:
Hi!

Facing issue when run taskiq

task_broker.py code below:
```
from taskiq_aio_pika import AioPikaBroker

broker = AioPikaBroker('amqp://guest:guest@localhost:5672')
```

deps:
```
taskiq                      0.10.0 
taskiq-aio-pika             0.4.0
taskiq-dependencies         1.4.0 
taskiq-redis                0.5.0 

```

Run configuration:

`taskiq worker core.taskiq.task_broker:broker`

Full error: 
```
[2023-10-19 12:31:01,682][taskiq.receiver.receiver][INFO   ][worker-1] Listening started.
Process worker-1:
Traceback (most recent call last):
  File "/Users/pavelgarkin/.pyenv/versions/3.9.9/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/Users/pavelgarkin/.pyenv/versions/3.9.9/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/pavelgarkin/Library/Caches/pypoetry/virtualenvs/invoices-Q1Rkz41h-py3.9/lib/python3.9/site-packages/taskiq/cli/worker/run.py", line 145, in start_listen
    loop.run_until_complete(receiver.listen())
  File "/Users/pavelgarkin/.pyenv/versions/3.9.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/pavelgarkin/Library/Caches/pypoetry/virtualenvs/invoices-Q1Rkz41h-py3.9/lib/python3.9/site-packages/taskiq/receiver/receiver.py", line 308, in listen
    gr.start_soon(self.runner, queue)
  File "/Users/pavelgarkin/Library/Caches/pypoetry/virtualenvs/invoices-Q1Rkz41h-py3.9/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 597, in __aexit__
    raise exceptions[0]
  File "/Users/pavelgarkin/Library/Caches/pypoetry/virtualenvs/invoices-Q1Rkz41h-py3.9/lib/python3.9/site-packages/taskiq/receiver/receiver.py", line 326, in prefetcher
    await self.sem_prefetch.acquire()
[...]

PR #270, opened 2023-12-28
Set uvloop for current thread explicitly.

ivaaahn:
Should help fix a similar problem to #229, but in the case of using uvloop.

I caught this problem on python3.8 with uvloop.


s3rius (approved):
Looks good to me. Thanks for contribution.

Issue #281, opened 2024-01-30
Reuse TypeAdapter instance to improve parsing params performance


unights:
According to the pydantic v2 [document](https://docs.pydantic.dev/latest/concepts/performance/#typeadapter-instantiated-once), the `TypeAdapter` should construct only once and reuse it.

Here is a simple test.

```python
from datetime import date, datetime
from timeit import timeit

from pydantic import TypeAdapter


def parse_obj(annot, value):
    ta = TypeAdapter(annot)

    result1 = timeit(lambda: ta.validate_python(value), number=10000)
    result2 = timeit(lambda: TypeAdapter(annot).validate_python(value), number=10000)

    print(f'{annot.__name__:<8} {result2 / result1:.2f}')


if __name__ == '__main__':
    for t, v in (
        (int, "0"),
        (float, '0.5'),
        (str, b"123"),
        (bool, 'true'),
        (date, '2024-01-01'),
        (datetime, '2024-01-01T00:00:00'),
    ):
        parse_obj(t, v)
```

output is

```plain
int      225.59
float    140.84
str      133.88
bool     228.47
date     357.20
datetime 319.76
```

Now the params parser in taskiq is 

```python
# taskiq/compat.py
def parse_obj_as(annot: T, obj: Any) -> T:
    return pydantic.TypeAdapter(annot).validate_python(obj)
```

And small changes may improve its performance

```python
from functools import lru_cache

@lru_cache
def get_adapter(annot: T) -> pydantic.TypeAdapter:
    return pydantic.TypeAdapter(annot)

def parse_obj_as(annot: T, obj: Any) -> T:
    return get_adapter(annot).validate_python(obj)
```

vvanglro:
Great! PR welcome.

PR #284, opened 2024-02-02
feat: reuse pydantic v2 TypeAdapter instance

unights:
Solves issue #281

### DIFF
--- a/taskiq/compat.py
+++ b/taskiq/compat.py
@@ -1,4 +1,5 @@
 # flake8: noqa
+from functools import lru_cache
 from typing import Any, Dict, Optional, Type, TypeVar, Union
 
 import pydantic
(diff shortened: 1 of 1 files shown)


taskiq/compat.py
```diff
@@ -13,8 +14,12 @@
 if IS_PYDANTIC2:
     T = TypeVar("T")
 
+    @lru_cache(maxsize=None)
```
vvanglro, line 17: Don't set it to None, which in long-running task libraries means that the cache will grow indefinitely.

taskiq/compat.py
```diff
@@ -13,8 +14,12 @@
...
+    def create_type_adapter(annot: T) -> pydantic.TypeAdapter:
```
s3rius, line 18: Should be

```suggestion
    def create_type_adapter(annot: T) -> pydantic.TypeAdapter[T]:
```

s3rius (changes requested):

taskiq/compat.py
unights: Oh yeah, I forgot it. Thanks for pointing out it.

taskiq/compat.py
unights: I thought the number of parameter types is usually limited. But this is indeed a dangerous action. Should we set it to the default 128, or a large value?

vvanglro: I suggest default.

s3rius (approved):

s3rius:
Thank you very much for your contribution.

PR #306, opened 2024-03-23
add log collection

zmarffy:
Implements a log collection system based on async task name.


taskiq/receiver/receiver.py
```diff
@@ -294,7 +313,7 @@ async def run_task(  # noqa: C901, PLR0912, PLR0915
         # Assemble result.
         result: "TaskiqResult[Any]" = TaskiqResult(
             is_err=found_exception is not None,
-            log=None,
+            log=log,
```
vvanglro, line 316: The log here seems to be just the normal log in the task function, do we need to collect the exceptions generated by `NoResultError` and `BaseException` as well?

zmarffy:
I have realized that this is not useful. It will only collect logs from the `taskiq.tasklogger` logger, not any modules that it calls. Celery essentially does the same thing with its `get_task_logger` function. However, Celery also implements something you can use to set up loggers to get imported modules to log with a Celery task ID: `after_setup_logger.connect` in conjunction with [`get_current_task`](https://github.com/celery/celery/blob/46992a11afe95e4149038406f456678fca6979d2/celery/_state.py#L121). Maybe we should look into implementing something like this for Taskiq. I have no clue how that would be possible in an async environment, but it would unlock some good potential.

PR #317, opened 2024-04-20
Change message acknowledgement

sfendourakis:
This proposed solution addresses an issue encountered when attempting to utilize acknowledgment functionality.

Currently, the `Receiver` module contains two types of messages: `AckableMessage` originating from the broker, which includes an `ack` method to be executed upon task completion for acknowledging the message, and `TaskiqMessage` containing execution information passed to the `run_task` method responsible for task execution.

The challenge arises when a task encounters an exception, as there is presently no mechanism to modify the behavior of the `ack` method, resulting in automatic acknowledgment of the message. To mitigate this, we have introduced the passing of the `ack` method within the `TaskiqMessage`. This grants access to the `ack` method within the `run_task` context, enabling us to utilize a `TaskiqMiddleware.on_error` method. This method accepts the `TaskiqMessage` and adjusts the behavior of the `ack` method according to the desired logic in the event of a task failure.

Returning to the `Receiver`, the updated `ack` method will now be executed, resolving the issue of unintended acknowledgments for failed tasks.

While this solution addresses the immediate challenge, it's important to recognize that there may be alternative approaches. I welcome any feedback, suggestions, or alternative solutions.

### DIFF
--- a/taskiq/message.py
+++ b/taskiq/message.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Awaitable, Callable, Dict, List, Optional, Union
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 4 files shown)


s3rius:
Generally acknowledgements are only for delivery. Means that once the message is delivered, it should be acknowledged. For error handling another mechanism should be used. For example retry middleware.

PR #337, opened 2024-06-16
Update progress_tracker.py

joeygrable94:
This allows pydantic 2.7 to pass core schema generation when no default is provided to TaskProgress class.


s3rius:
Seems good to me.

s3rius:
Thanks.

PR #343, opened 2024-07-19
fix: sync decorators

Sobes76rus:
use case
```[python]
@broker.task()
@validate_call()
async def task(arg1: str) -> None: ...
```

so, the thing is that pydantic's `validator_call` is syncronious decorator which returns awaitable future after call
right now receiver can't work with sync decorators on async functions

### DIFF
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -267,8 +267,16 @@ class Receiver:
             if timeout is not None:
                 if not is_coroutine:
                     logger.warning("Timeouts for sync tasks don't work in python well.")
-                target_future = asyncio.wait_for(target_future, float(timeout))
-            returned = await target_future
+
+                with anyio.fail_after(float(timeout)):
+                    while inspect.isawaitable(target_future):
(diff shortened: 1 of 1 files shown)


taskiq/receiver/receiver.py
```diff
@@ -267,8 +267,16 @@ async def run_task(  # noqa: C901, PLR0912, PLR0915
             if timeout is not None:
                 if not is_coroutine:
                     logger.warning("Timeouts for sync tasks don't work in python well.")
-                target_future = asyncio.wait_for(target_future, float(timeout))
-            returned = await target_future
+
+                with anyio.fail_after(float(timeout)):
+                    while inspect.isawaitable(target_future):
```
s3rius, line 272: Why do we need this while loop? Won't it return an awaitable after first iteration? Can we just check it with if?

Sobes76rus: done

Sobes76rus:
@s3rius all done as you suggested

s3rius (approved):

PR #349, opened 2024-08-18
Add factories

dima-dmytruk23:


taskiq/cli/scheduler/args.py
```diff
@@ -32,7 +34,21 @@ def from_cli(cls, args: Optional[Sequence[str]] = None) -> "SchedulerArgs":
             formatter_class=ArgumentDefaultsHelpFormatter,
             description="Subcommand to run scheduler",
         )
-        parser.add_argument("scheduler", help="Path to scheduler")
+        parser.add_argument(
```
s3rius, line 37: We don't need two parameters. Specify only factory. We can check if the imported scheduler is a factory after we import it.

taskiq/abc/broker_factory.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+from taskiq.abc.broker import AsyncBroker
+
+
+class BrokerFactory(ABC):
```
s3rius, line 6: Why do we need a class for that? Shouldn't just path to function be enough?

taskiq/cli/worker/args.py
```diff
@@ -59,13 +60,24 @@ def from_cli(
         """
         parser = ArgumentParser(formatter_class=ArgumentDefaultsHelpFormatter)
         parser.add_argument(
-            "broker",
+            "--broker",
```
s3rius, line 63: Same question as for the scheduler.

taskiq/abc/scheduler_factory.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+from taskiq import TaskiqScheduler
+
+
+class TaskiqSchedulerFactory(ABC):
```
s3rius, line 6: Same question.

s3rius (changes requested):

dima-dmytruk23:
@s3rius fixed

Molandrious:
@s3rius can you merge this?

s3rius (approved):

s3rius (approved):

PR #357, opened 2024-08-29
Fix shutdown middleware hook

hawang-wish:
Hi,

Thank you for creating this awesome package!

I have spotted a potential typo issue that stops our shutdown middleware hook from working properly.

I have thought that perhaps I could add some tests, but I've then decided to make it in another PR just to reduce some of my work.

Please feel free to comment, and let me know in case you prefer to have tests for this PR.


s3rius (approved):
Yes. It seems like a potential bug. Thanks for noticing. Tests will be much appriciated, but you can add them in a separate request.

PR #358, opened 2024-08-29
Fix InMemoryBroker to use startup and shutdown middleware hooks

hawang-wish:
Hi,

This is the second patch I would like to introduce. Previously, we overrode the `startup` and `shutdown` to ensure both `client` and `worker` events would be handled properly. However, this made InMemoryBroker unable to handle `startup` and `shutdown` middleware hooks.

This change brings the handling of `startup` and `shutdown` middleware hooks back. Feel free to let me know if you have any concerns or suggestions!

### DIFF
--- a/taskiq/brokers/inmemory_broker.py
+++ b/taskiq/brokers/inmemory_broker.py
@@ -173,13 +173,23 @@ class InMemoryBroker(AsyncBroker):
 
     async def startup(self) -> None:
         """Runs startup events for client and worker side."""
-        for event in (TaskiqEvents.CLIENT_STARTUP, TaskiqEvents.WORKER_STARTUP):
-            for handler in self.event_handlers.get(event, []):
-                await maybe_awaitable(handler(self.state))
+        for handler in self.event_handlers.get(
+            TaskiqEvents.CLIENT_STARTUP
+            if self.is_worker_process
+            else TaskiqEvents.WORKER_STARTUP,
+            [],
+        ):
+            await maybe_awaitable(handler(self.state))
+        await super().startup()
 
     async def shutdown(self) -> None:
         """Runs shutdown events for client and worker side."""
-        for event in (TaskiqEvents.CLIENT_SHUTDOWN, TaskiqEvents.WORKER_SHUTDOWN):
-            for handler in self.event_handlers.get(event, []):
-                await maybe_awaitable(handler(self.state))
+        for handler in self.event_handlers.get(
+            TaskiqEvents.CLIENT_SHUTDOWN
+            if self.is_worker_process
+            else TaskiqEvents.WORKER_SHUTDOWN,
+            [],
+        ):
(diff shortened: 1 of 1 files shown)


s3rius:
Thanks for contribution. I don't think that this change is necessary, because Inmemory broker cannot listen to tasks. So `is_worker_process` variable will never be set to True. Can you please describe the case when this might be found useful?

taskiq/brokers/inmemory_broker.py
```diff
@@ -173,13 +173,23 @@ def listen(self) -> AsyncGenerator[bytes, None]:
...
         """Runs startup events for client and worker side."""
-        for event in (TaskiqEvents.CLIENT_STARTUP, TaskiqEvents.WORKER_STARTUP):
-            for handler in self.event_handlers.get(event, []):
-                await maybe_awaitable(handler(self.state))
+        for handler in self.event_handlers.get(
+            TaskiqEvents.CLIENT_STARTUP
+            if self.is_worker_process
+            else TaskiqEvents.WORKER_STARTUP,
+            [],
+        ):
```
s3rius, line 181: Shouldn't it be reversed? 

```suggestion
        for handler in self.event_handlers.get(
            TaskiqEvents.WORKER_STARTUP
            if self.is_worker_process
            else TaskiqEvents.CLIENT_STARTUP,
            [],
        ):
```

taskiq/brokers/inmemory_broker.py
```diff
@@ -173,13 +173,23 @@ def listen(self) -> AsyncGenerator[bytes, None]:
...
 
     async def shutdown(self) -> None:
         """Runs shutdown events for client and worker side."""
-        for event in (TaskiqEvents.CLIENT_SHUTDOWN, TaskiqEvents.WORKER_SHUTDOWN):
-            for handler in self.event_handlers.get(event, []):
-                await maybe_awaitable(handler(self.state))
+        for handler in self.event_handlers.get(
+            TaskiqEvents.CLIENT_SHUTDOWN
+            if self.is_worker_process
+            else TaskiqEvents.WORKER_SHUTDOWN,
```
s3rius, line 190: This one as well.

```suggestion
            TaskiqEvents.WORKER_SHUTDOWN
            if self.is_worker_process
            else TaskiqEvents.CLIENT_SHUTDOWN,
```

s3rius (changes requested):
The request looks good to me. But I have a small question.

taskiq/brokers/inmemory_broker.py
qodize: @s3rius No, it should not be reversed, because then it calls super().startup(). This was made to handle both CLIENT_STARTUP and WORKER_STARTUP events, one in InMemoryBroker.startup() and other in AsyncBroker.startup(). The same applies to shutdown.

PR #359, opened 2024-08-29
Invert execution orders of post, on_error and shutdown middleware hooks

hawang-wish:
Hi,

I've further noticed that we could introduce a slight improvement to make middlewares follow the "onion" execution order. For example, if a `message` is wrapped with `pre_send` of multiple middlewares, it's generally better to run `post_send` in reversed order.

I've then inverted the order of `post_execute`, `shutdown`, `on_error`, and `post_save` so that all existing middleware hooks have an onion-ish invoke order.

A practical use case is that some of my middlewares do modifications to add a mark in `pre_send` and remove it in `post_send`. However, without this improvement I couldn't make the mark invisible to other middlewares.

The unit tests I added could fail because of other issues: https://github.com/taskiq-python/taskiq/pull/358 and https://github.com/taskiq-python/taskiq/pull/357 

As always, let me know if you have suggestions or concerns, and thanks for the review in advance!

### DIFF
--- a/taskiq/abc/broker.py
+++ b/taskiq/abc/broker.py
@@ -213,7 +213,7 @@ class AsyncBroker(ABC):
         for handler in self.event_handlers[event]:
             await maybe_awaitable(handler(self.state))
 
-        for middleware in self.middlewares:
+        for middleware in reversed(self.middlewares):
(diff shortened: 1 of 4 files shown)


s3rius:
Please fix lints and tests before the review.

PR #360, opened 2024-08-29
Added on_send_error middleware hook

hawang-wish:
Hi,

I am currently working on a retrying feature to ensure the delivery of messages when the queue service is unstable. I found that it could be really helpful to have this `on_send_error` so that we can store the failed message and retry later.

I've also added a `bool` return to make it able to mute exceptions so that we don't have to try-catch the kicking everywhere.

The execution order of `on_send_error` is reversed, just like what I suggested in https://github.com/taskiq-python/taskiq/pull/359 .

Please let me know if you have some ideas, suggestions or concerns, and big thanks in advance!

### DIFF
--- a/taskiq/abc/middleware.py
+++ b/taskiq/abc/middleware.py
@@ -131,7 +131,7 @@ class TaskiqMiddleware:  # pragma: no cover
         self,
         message: "TaskiqMessage",
         broker_message: "BrokerMessage",
-        exception: BaseException,
+        exception: Exception,
     ) -> "Union[Union[bool, None], Coroutine[Any, Any, Union[bool, None]]]":
         """
         This function is called when exception is raised while sending a message.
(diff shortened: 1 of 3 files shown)


s3rius (approved):
Thanks for your contributions. I like the idea.

s3rius:
Can you please doc entry about this functionality in docs folder?

taskiq/abc/middleware.py
```diff
@@ -126,3 +126,25 @@ def on_error(
         :param result: returned value.
         :param exception: found exception.
         """
+
+    def on_send_error(
+        self,
+        message: "TaskiqMessage",
+        broker_message: "BrokerMessage",
+        exception: Exception,
+    ) -> "Union[Union[bool, None], Coroutine[Any, Any, Union[bool, None]]]":
```
s3rius, line 135: This better be 
```suggestion
    ) -> "Union[Optional[bool], Coroutine[Any, Any, Optional[bool]]]":
```

s3rius (approved):

PR #362, opened 2024-09-04
fix: SIGINT worker processes when process_manager starts

sminnee:
Without this change, I found that workers could get stuck on a long-running task. Worker processes are sent a SIGINT, and then that signal is handled appropriately to exit the prefetcher & runner.

Is this appropriate as general purpose behaviour, and if not, would it make sense as a command-line option?

### DIFF
--- a/taskiq/cli/worker/process_manager.py
+++ b/taskiq/cli/worker/process_manager.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import signal
 import sys
 from contextlib import suppress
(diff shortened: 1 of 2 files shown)


sminnee:
I note that tasks cancelled in this way have their result returned as error = KeyboardInterrupt.

In my own application, I have a middleware that re-kicks these tasks when the worker next starts.

s3rius (approved):
Makes sense to me. Thanks for the contribution. Will be released soon.

Issue #322, opened 2024-04-30
Add messages rejection


vasilbekk:
Thank you so much for creating such a cool library, but I found a problem.

rabbitmq has support for dead messages, and taskiq-aio-pika even creates a queue for them.

But in the Receiver implementation in tasker, there is only message.ack(), and there is no option for message.reject()

And in Receiver.callback(), ack() is always called even if the task failed with an error

I rewrote the class a little bit, and added the optional reject() callable field to AckableMessage (I called RejectableMessage), also rewrote the broker so that in listen() it returns RejectableMessage instead of AckableMessage, and rewrote `Receiver.callback() so that in the case of result.is_err it calls reject(), and not ack()

I could clean up a bit and send a PR with message rejection implementation (especially useful in rabbitmq) if needed.

It would also be interesting to get an answer to 2 questions
1. Why is the creation of a queue for dead messages implemented in taskiq-aio-pika, but it is essentially always empty, was it laid down for the future or broke?

2. Perhaps there is another implementation of repeating error messages and I do not see it, I would be glad to hear suggestions

a1d4r:
Hello, Sergey! I've encountered the same problem. I would rather expect sending failing tasks to the dead letter queue rather than ack'ing them. Could you please share a draft PR or a gist? I'm pretty sure it might help many people including me.

xg-wang:
Hi, I'm thinking about the same rejection/retry on failed execution. For my use case I use SQS as the broker, the SQS approach is I need to explicitly delete the message receipt handle on success, otherwise SQS makes the message visible again for execution after timeout. Therefore the best rejection/retry is to do nothing on error.

Is it possible to pass result to AckableMessage.ack()? This way broker can check whether result is error to decide how to ack.

PR #371, opened 2024-10-15
Do not ACK errored messages

danngalann:
I had an issue with my FastAPI project where a task was throwing an exception and yet the message was being acknowledged and lost forever. I tried using the context inside the task to reject the message as shown [here](https://taskiq-python.github.io/guide/architecture-overview.html#context), but the task seemed to hang while trying to resolve the context dependency. I ended up implementing a fix I'll describe below.

For reference, here's a simplified version of my code:
```python
# broker.py
import taskiq_fastapi
from taskiq_aio_pika import AioPikaBroker

from settings import settings

broker = AioPikaBroker(
    settings.rabbit_dsn,
    queue_name=f"myapp_{settings.environment}",
    exchange_name=f"myapp_{settings.environment}",
    dead_letter_queue_name=f"myapp_{settings.environment}.dead_letter",
    delay_queue_name=f"myapp_{settings.environment}.delay",
    qos=1,  # Consume one message at a time
)
```

```python
# myservice.py
@broker.task
async def store(
    task_data,
) -> None:
    service = SomeService()
    data = await service.fetch_data_rpc(
        task_data
    )
```

Here, the `fetch_data_rpc` inside my task throws an exception. I'd expect the message to be requeued, sent to the dead letter configured in the broker, or at least not acknowledged, but that is not the case.

I'm not sure if this is an issue or an intended behavior, as I can see that in your `receiver.py` all messages are acknowledged regardless of their errors.
```python
if self.ack_time == AcknowledgeType.WHEN_RECEIVED and isinstance(
    message,
    AckableMessage,
):
    await maybe_awaitable(message.ack())
```

This may be related with [this issue](https://github.com/taskiq-python/taskiq/issues/322),

In the end, my solution was to wrap the ACK in an if that checks if the message had errors. Now, if a task fails, the message is not acknowledged. This solves my issue, as failed messages are not lost now.

I'm sending my fix as a PR in case you consider it useful for use cases other than mine.

### DIFF
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -144,7 +144,15 @@ class Receiver:
             message,
             AckableMessage,
         ):
-            await maybe_awaitable(message.ack())
+            if not result.is_err:
+                await maybe_awaitable(message.ack())
+            else:
+                # Do not acknowledge the message
+                logger.error(
+                    "Task %s with ID %s failed. Message not acknowledged.",
+                    taskiq_msg.task_name,
+                    taskiq_msg.task_id,
+                )
 
         result = await self.run_task(
             target=task.original_func,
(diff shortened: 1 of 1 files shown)


Sobes76rus:
Yes, it is intended behavior. 
If you wish to retry on error, u could implement `RetryMiddleware` for `AioPikaBroker` or define your own `ack` method for message
It is problem with `AioPikaBroker` and not `Receiver`

s3rius:
I agree with @Sobes76rus. The thing is that ack mechanism is intended to check on connection issues. For example if during reading message worker went down and couldn't process the message it would be given to another worker. But if message was received then it's up to the application to handle exceptions.

Issue #380, opened 2024-11-20
Graceful Shutdown Issue with Taskiq in K8s


Bohdan-Ilchyshyn:
I am using Taskiq with Kubernetes and KEDA scaling, and I am encountering issues with implementing a graceful shutdown process for my workers.

## Problem Description

When a pod receives a SIGTERM signal:
* I need to stop the worker from receiving new tasks.
* The worker should wait for currently running tasks to complete before shutting down.

I have configured the following settings:

    shutdown-timeout = 70
    wait-tasks-timeout = 60

However, this does not work as expected because wait-tasks-timeout is only used when the listener receives the QUEUE_DONE message. The QUEUE_DONE message is sent exclusively in the Listener.prefetcher.
## Expected Behavior

Upon receiving a SIGTERM signal, the worker should:

* Immediately stop accepting new tasks.
* Wait for all in-progress tasks to complete within the wait-tasks-timeout duration.

## Questions

* How can I properly implement this behavior for a graceful shutdown?
* Are there existing hooks or configurations I should use to address this issue?
* Should QUEUE_DONE messages be triggered elsewhere to support this scenario?

s3rius:
This sounds like a serious bug. I'll look into it.

Bohdan-Ilchyshyn:
I started working on my own solution yesterday, but since you've already completed yours, I’ll wait for the new version and then ask a few questions. Perhaps I don’t fully understand why the following things can't be done:

* Why can't we move the task registry into __init__?
* Why can't we start two tasks and store them in __init__?
* Why can't we create a shutdown method to terminate the tasks?

For example, I was thinking of doing it this way:

### Receiver
```python
    def __init__(...):
        ...
        self._allow_pick_jobs: bool = False
        self._tasks: Set[asyncio.Task[Any]] = set()

        self.prefetcher_task: Task | None = None
        self.runner_task: Task | None = None

    def _task_cb(self, task: "asyncio.Task[Any]") -> None:
        self._tasks.discard(task)
        if self.sem is not None:
            self.sem.release()

    async def _sleep_until_tasks_complete(self) -> None:
        while len(self._tasks) != 0:
            await asyncio.sleep(0.1)

    async def wait_for_tasks_to_complete(self) -> None:
        if self.prefetcher_task is not None:
            self.prefetcher_task.cancel()

        with contextlib.suppress(asyncio.TimeoutError):
            await asyncio.wait_for(
                self._sleep_until_tasks_complete(),
[...]

Issue #325, opened 2024-05-16
Graceful shutdown and reload


dima-kov:
Hi! Thank you for to simple and easy to use tasks manager lib!

Please help me to understand, how to achieve graceful reload? Can't find any docs about this. 

**How it works now**: when I cancel the process KeybordInterrupt is raised and tasks are cancelled (unfinished). 
**What I/m looking for:** possibility to send HUP event, like I use for uvicorn servers manged by supervisor. How it works:
1. stops accepting new requests;
2. finishes all open requests;
3. restarts the process;
4. accepts new requests;

What is an expected way for taskiq to handle code updates (releases)?

s3rius:
I guess graceful updates are not supported right now, the main process won't handle a HUP signal, but it can be added.

I think it's a good thing to have.

dima-kov:
Wow, that's a really big problem when required to use in production.

What can I do to make it happen? How can I contribute?

s3rius:
Should be easy to add. Basically, we have everything for this feature to be implemented. 

In the [ProcessManager](https://github.com/taskiq-python/taskiq/blob/4ee831214ba67172096e64c78f2f1fb13d018b39/taskiq/cli/worker/process_manager.py#L143) you can define a function that reloads all subprocesses. 

```python
def reload_all(self):
	self.action_queue.put(ReloadAllAction())
``` 

And somewhere [here](https://github.com/taskiq-python/taskiq/blob/4ee831214ba67172096e64c78f2f1fb13d018b39/taskiq/cli/worker/run.py#L191), before the process manager has started, you can define signal handler that will call a reload_all action on a `process_manager` in case of sigHUP.

s3rius:
Okay. Since this issue is a bit stale, I will try implementing it.

s3rius:
@itssimon, @dima-kov, please verify that the feature was implemented correctly. It was released in 0.11.4.

dima-kov:
Took a look on how it works.
Test plan is: send two long async tasks, and during execution send HUP signal. 
**Expected**: existing tasks should be finished succesfull.
**Actual**: worker reloaded, just same like before, stopping the task execution.

LOGS:
```
[2024-07-04 10:37:15,258][taskiq.receiver.receiver][INFO   ][worker-0] Listening started.
[...]

PR #381, opened 2024-11-21
Added correct graceful shutdown.

s3rius:
Fixes https://github.com/taskiq-python/taskiq/issues/380 
Fixes https://github.com/taskiq-python/taskiq/issues/325

This PR rewrites graceful shutdowns completely and now it should be able to gracefully wait for all tasks to complete and implement proper reloading.

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -86,12 +86,25 @@ when you modify ignored files. To disable this functionality pass `--do-not-use-
 
 ### Graceful reload (available only on Unix systems)
 
-To perform graceful reload, send `SIGHUP` signal to the main worker process. This action will reload all workers with new code. It's useful for deployment that requires zero downtime, but don't use orchestration tools like Kubernetes.
+To perform graceful reload, send `SIGHUP` signal to the main worker process. This action will reload all workers with new code. It's useful for deployment that requires zero downtime, but without using heavy orchestration tools like Kubernetes.
+
 
 ```bash
 taskiq worker my_module:broker
 kill -HUP <main pid>
 ```
+
+### Graceful and force shutdowns
+
+If you send `SIGINT` or `SIGKILL` to the main process by pressing <kbd>Ctrl</kbd>+<kbd>C</kbd> or using the `kill` command, it will initiate the shutdown process.
+By default, it will stop fetching new messages immediately after receiving the signal but will wait for the completion of all currently executing tasks.
+
+If you don't want to wait too long for tasks to complete each time you shut down the worker, you can either send termination signals three times to the main process to perform a hard kill or configure the `--wait-tasks-timeout` to set a hard time limit for shutting down.
+
+::: tip Cool tip
+The number of signals before a hard kill can be configured with the `--hardkill-count` CLI argument.
+:::
+
 
 ### Other parameters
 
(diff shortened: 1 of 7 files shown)


remixer-dec:
This PR made it impossible to track `docker stop` shutdown events from within worker script, signals are just ignored

s3rius:
@remixer-dec, can you please clarify your situation a bit? Because it's not clear to me how you try to track exits.

remixer-dec:
@s3rius I had signal handling logic implemented in my worker script that saves the state when worker exits, it no longer works after I upgraded taskiq.
It was implemented this way:

```python
import signal
import logging

logger = logging.getLogger(__name__)
logger.info("Worker active")

def on_worker_shutdown():
  logger.info("Stop signal received")
  # other custom logic


signal.signal(signal.SIGINT, lambda *args, **kwargs: on_worker_shutdown())
signal.signal(signal.SIGTERM, lambda *args, **kwargs: on_worker_shutdown())
```

I'm trying to find other way to somehow catch the fact that the worker is being shutdown, `atexit` and signals doesn't work. 
Gemini suggested me to do this:
```python
@broker.on_event(TaskiqEvents.WORKER_SHUTDOWN)
async def on_worker_shutdown(state: TaskiqState) -> None:
  logger.info("Stop signal received")
  # other custom logic

# or like this
broker.add_event_handler(TaskiqEvents.WORKER_SHUTDOWN, on_worker_shutdown)
```
Both doesn't seem to work.

remixer-dec:
I downgraded and I'm still not getting intended behavior, maybe something else changed in the environment, I'll investigate further and will give more details once I find the reason, sorry for bothering you too soon.

s3rius:
Don't worry. Please write me once you got what was happening.

remixer-dec:
@s3rius I found the issue, the signals still work fine in the latest version of taskiq, the issue was in docker+bash signal delivery when a background process is provided in `command`. I recently added taskiq scheduler for cron tasks in the container, this caused the issue.

so in `docker-compose.yaml`
this works
```yaml 
  worker-service:
    # ...
    command: >
      bash -c 'taskiq worker worker-async:broker -w 1'
```

but this doesn't
```yaml
  worker-service:
    # ...
    command: >
      bash -c 'taskiq scheduler worker-async:scheduler & taskiq worker worker-async:broker -w 1'
```
and signals are not getting delivered to worker.  

The solution to this is quite simple, use exec before running worker:

```yaml
  worker-service:
    # ...
    command: >
      bash -c 'taskiq scheduler worker-async:scheduler & exec taskiq worker worker-async:broker -w 1'
```

s3rius:
I would suggest to check on https://github.com/taskiq-python/examples/blob/master/fastapi-app/docker-compose.yml. This config works flawlessly.

remixer-dec:
@s3rius oh, you have scheduler and worker in separate containers, maybe that's more stable in a long run, thanks!

s3rius:
Yes! Also it helps scaling. Because you can scale workers separately. 

Btw, our company has opensourced helm [charts](https://github.com/Intreecom/charts/tree/master/charts/py-app/templates) for py-application. And we have different deployments for [taskiq-scheduler.yaml](https://github.com/Intreecom/charts/blob/master/charts/py-app/templates/taskiq-scheduler.yaml) and [taskiq-worker.yaml](https://github.com/Intreecom/charts/blob/master/charts/py-app/templates/taskiq-worker.yaml).

s3rius:
@remixer-dec, please upgrade to version 0.11.14. We found were was the issue. It was really hard to reproduce, but it was there. 

Thanks for pointing it out. You can check what was the issue in the Pull Request.

PR #390, opened 2024-12-15
Enhance boolean label handling

### LINKED ISSUE #386: `AttributeError: 'bool' object has no attribute 'lower'`
Hello @s3rius! Thank you for creating taskiq.

I encountered an exception (something like `AttributeError: 'bool' object has no attribute 'lower'`) with `taskiq` when attempting to use my custom `SimpleRetryMiddleware`. The problem occurred with the following code:

```python
@taskiq_broker.task(retry_on_error=True)
async def _process_not_reached_deal(deal_details: BitrixDealResultData) -> Optional[bool]:
```

Unfortunately, I don't have the full traceback of the exception to provide for troubleshooting.

I've implemented a fix, and the code is now working correctly:
```python
_LABEL_PARSERS: Dict[LabelType, Callable[[str], Any]] = {
    LabelType.INT: int,
    LabelType.STR: str,
    LabelType.FLOAT: float,
    LabelType.BOOL: lambda x: str(x).lower() == "true",
    LabelType.BYTES: base64.b64decode,
    LabelType.ANY: lambda x: x,
}
```
However, I'm not entirely sure if the boolean value conversion logic (lambda x: str(x).lower() == "true") is the most appropriate approach. I'd appreciate your thoughts on this implementation.

ZhymabekRoman:
Fixes #386

### DIFF
--- a/taskiq/labels.py
+++ b/taskiq/labels.py
@@ -18,7 +18,7 @@ _LABEL_PARSERS: Dict[LabelType, Callable[[str], Any]] = {
     LabelType.INT: int,
     LabelType.STR: str,
     LabelType.FLOAT: float,
-    LabelType.BOOL: lambda x: x.lower() == "true",
+    LabelType.BOOL: lambda x: str(x).lower() == "true",
     LabelType.BYTES: base64.b64decode,
     LabelType.ANY: lambda x: x,
 }


s3rius (approved):

s3rius:
Can you rebase onto last commit from master branch?

I have updated a workflow, so tests should run after the update.

ZhymabekRoman:
I didn't see it. I apologise.

PR #394, opened 2024-12-22
Update deps

sobolevn:
I found out that there were several problems:
- 3.8 is not supported
- 3.12 and 3.13 metadata was missing
- `[tool.poetry.dev-dependencies]` is deprecated


s3rius (approved):

s3rius:
Thanks for the PR. Closed as fixed in #413.

PR #395, opened 2024-12-28
Add max_attempts_at_message

Arseniy-Popov:
This addresses the issue https://github.com/taskiq-python/taskiq-aio-pika/issues/35 by adding support for a limit on the number of attempts to process the same message.


s3rius:
I don't think this PR should be implemented this way and affect base classes such as AckableMessage. Also in here I don't see anywhere how you set this redeliver information to a message. If you expect brokers to do it, then I refuse to merge it, because Broker API should remain simple to implement. 

I think that it will be much easier for you to implement such functionality as a middleware and add it in [taskiq.middlewares](https://github.com/taskiq-python/taskiq/tree/master/taskiq/middlewares).

Arseniy-Popov:
Thanks for the reply. 
> I don't see anywhere how you set this redeliver information to a message. 

As my issue explains, RabbitMQ does this by itself: if a message that was delivered isn't acked but the connection with the consuming client closes, RabbitMQ requeues the message to the front of the queue and sets a `x-delivery-count` header.
> base classes such as AckableMessage 

The issue currently is that message headers, including the one talked about above, aren't being passed to the message the receiver works with. They need to be put somewhere, either into the `TaskiqMessage` or its wrapper.
> implement such functionality as a middleware 

The middleware would still need access to the message headers. It would also need access to the message's ack callback, and the `pre_execute` method would need to be able to break the current processing (raising an exception in it doesn't seem viable at first glance).

Issue #348, opened 2024-08-17
Will intervals be added?


kanewi11:
How can I implement launching tasks with an interval of less than one minute, for example 10 seconds?

This is necessary in order not to create:
```
while True:
    await asyncio.sleep(10)
    ...
```
And do not wrap it in `try: ... except ExampleException: ...`, so that the task does not crash and does not process the code anymore.

Issue #290, opened 2024-02-14
Schedule Tasks with Intervals shorter Than 1 Minute


Soures888:
Hello! I need the ability to run tasks at intervals shorter than one minute. Currently, the scheduler seems to run tasks every minute, often with some delays. Are there examples or methods available that would allow me to run tasks at shorter intervals, such as every 20 seconds?

vadympop:
Also interested in this functionality

danfimov:
This functionality has already been implemented in https://github.com/taskiq-python/taskiq/pull/498 and release in [0.11.17 version](https://github.com/taskiq-python/taskiq/releases/tag/0.11.17).

You can check updated docs about task scheduling [here](https://taskiq-python.github.io/guide/scheduling-tasks.html).

PR #399, opened 2025-01-11
Add intervals for scheduled task

so-saf:
#290 #348

Hi! First, thank you for the project!

In this PR, I want to add intervals for scheduled tasks. I am attaching one commit as a draft to coordinate further changes with you.
I have implemented MVP, and a lot of related code has yet to be written, including documentation and tests.

The implementation is quite simple:
 We go through all scheduled tasks once a second and send them to the broker.
- For cron tasks, the frequency remained around a minute.

Of the additional advantages, it can be noted that tasks marked with the "time" label are scheduled with less delay.

What do you think?

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -21,18 +21,18 @@ We created this project because we couldn't find any project that can execute an
 
 You might have seen projects built on top of asyncio that solve a similar problem, but here's a comparison table of the taskiq and other projects.
 
-|                Feature name | Taskiq |  Arq  | AioTasks |
-| --------------------------: | :----: | :---: | :------: |
-|         Actively maintained |   ✅    |   ✅   |    ❌     |
-|    Multiple broker backends |   ✅    |   ❌   |    ✅     |
-|    Multiple result backends |   ✅    |   ❌   |    ❌     |
-|  Have a rich documentation  |   ✅    |   ❌   |    ❌     |
-|   Startup & Shutdown events |   ✅    |   ✅   |    ❌     |
-| Have ability to abort tasks |   ❌    |   ✅   |    ❌     |
-|          Custom serializers |   ✅    |   ✅   |    ❌     |
-|        Dependency injection |   ✅    |   ❌   |    ❌     |
-|              Task pipelines |   ✅    |   ✅   |    ❌     |
-|              Task schedules |   ✅    |   ✅   |    ❌     |
-|          Global middlewares |   ✅    |   ❌   |    ❌     |
+|                Feature name | Taskiq |  Arq  | AioTasks | streaQ |
+| --------------------------: | :----: | :---: | :------: | :----: |
+|         Actively maintained |   ✅    |   ✅   |    ❌     |   ✅   |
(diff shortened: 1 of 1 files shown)


Mothilal-hire10x:
> #290 #348
> 
> Hi! First, thank you for the project!
> 
> In this PR, I want to add intervals for scheduled tasks. I am attaching one commit as a draft to coordinate further changes with you. I have implemented MVP, and a lot of related code has yet to be written, including documentation and tests.
> 
> The implementation is quite simple: We go through all scheduled tasks once a second and send them to the broker.
> 
> * For cron tasks, the frequency remained around a minute.
> 
> Of the additional advantages, it can be noted that tasks marked with the "time" label are scheduled with less delay.
> 
> What do you think?

Need to be merge

taskiq/cli/scheduler/run.py
```diff
@@ -142,32 +149,53 @@ async def run_scheduler_loop(scheduler: TaskiqScheduler) -> None:
...
+                to_send = False
+                if task.time is not None and now >= task.time:
+                    to_send = True
+                elif task.interval is not None:
+                    last_call = interval_tasks_last_run.get(task.task_name)
+                    if (
+                        last_call is not None
+                        and round((now - last_call).total_seconds()) < task.interval
+                    ):
+                        to_send = False
```
s3rius, line 175: `to_send` is `False` by default. You can negate this if condition and only leave the second branch.

taskiq/cli/scheduler/run.py
```diff
@@ -142,32 +149,53 @@ async def run_scheduler_loop(scheduler: TaskiqScheduler) -> None:
...
+                    else:
+                        to_send = True
+                        interval_tasks_last_run[task.task_name] = now
+                elif now > next_minute and task.cron is not None:
```
s3rius, line 179: I'm not sure this condition will work as expected, and most likely, it won't.

Currently, the `now` value is updated on each iteration of the while loop, and so is `next_minute`. This could cause an issue, especially if it takes slightly longer than 1 second to obtain all scheduled tasks due to factors like network lag. In that case, there would be no delay between setting the `next_minute` and updating the `now` value.

s3rius: Easier solution would be to have a dict of last cron runs and remember which tasks have run in this very minute. Similarly as you've done with interval_tasks.

taskiq/cli/scheduler/run.py
```diff
@@ -142,32 +149,53 @@ async def run_scheduler_loop(scheduler: TaskiqScheduler) -> None:
     """
     loop = asyncio.get_event_loop()
     running_schedules = set()
+
+    interval_tasks_last_run: "Dict[str, datetime]" = {}
```
s3rius, line 153: Since it's a `dict[str, datetime]` it seems like if you are going to have multiple intervals, you solution will most probably crumble apart. You need to have a nested list (because intervals can have the same value). Each list item will be an interval value and last run.

s3rius (changes requested):

s3rius:
Also, please rebase your branch onto master. There were some changes.

PR #404, opened 2025-02-04
Fix task delay for specific time with milliseconds

megalaren:
When using the `.schedule_by_time()` method, I encountered a problem where the task starts more than 1 second earlier than the scheduled time. I believe the scheduler should not start tasks before the specified time.
This PR fixes this issue. Also added a test that simulates my problem.


s3rius (approved):

s3rius:
LGTM. Will be released soon.

PR #413, opened 2025-02-26
Fixed type inference for new versions of pyright and dependencies were updated.

s3rius:
This PR fixes type issues introduced by a new pyright update. 

In https://github.com/microsoft/pyright/releases/tag/1.1.394 they have made a change that breaks return_type inference for kicked tasks:

> Changed the type of an async def function from typing.Coroutine to types.CoroutineType for improved type accuracy.

This PR addresses this behavior and now return_value of the result shows correct types.


chandr-andr (approved):
Awesome PR.

PR #423, opened 2025-03-12
Fix Typos, Warn on Unmatching Brokers

TheTechromancer:
Hi all, thanks for making this repo; it's really useful.

This PR fixes some typos and also adds a warning when a scheduled task's broker doesn't match the broker it's being executed on. This has fucked me over several times when scheduled tasks are registered but silently fail to execute. This was my own fault since I was double-registering tasks. But it's a nasty pitfall.

Should we detect mismatched brokers a bit sooner, i.e. when the task is first registered? Probably if a user tries to register a function that's already an instance of a decorated task, we should throw an error if the brokers don't match.

### DIFF
--- a/taskiq/abc/broker.py
+++ b/taskiq/abc/broker.py
@@ -29,6 +29,7 @@ from taskiq.abc.serializer import TaskiqSerializer
 from taskiq.acks import AckableMessage
 from taskiq.decor import AsyncTaskiqDecoratedTask
 from taskiq.events import TaskiqEvents
+from taskiq.exceptions import TaskRejectedError
 from taskiq.formatters.proxy_formatter import ProxyFormatter
 from taskiq.message import BrokerMessage
 from taskiq.result_backends.dummy import DummyResultBackend
@@ -517,7 +518,13 @@ class AsyncBroker(ABC):
         By default we register tasks in local task registry.
         But this behaviour can be changed in subclasses.
 
+        This method may raise TaskRejectedError if task has already been registered to a different broker.
+
         :param task_name: Name of a task.
         :param task: Decorated task.
         """
+        if task.broker != self:
+            raise TaskRejectedError(
+                f"Task already has a different broker ({task.broker})",
+            )
         self.local_task_registry[task_name] = task
--- a/taskiq/abc/broker.py
+++ b/taskiq/abc/broker.py
@@ -29,7 +29,7 @@ from taskiq.abc.serializer import TaskiqSerializer
 from taskiq.acks import AckableMessage
 from taskiq.decor import AsyncTaskiqDecoratedTask
 from taskiq.events import TaskiqEvents
-from taskiq.exceptions import TaskRejectedError
+from taskiq.exceptions import TaskBrokerMismatchError
 from taskiq.formatters.proxy_formatter import ProxyFormatter
 from taskiq.message import BrokerMessage
 from taskiq.result_backends.dummy import DummyResultBackend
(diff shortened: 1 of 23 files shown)


s3rius:
Yooooo. That's really bad. Sorry for not thinking of this case. I should have created this warning earlier. 

This PR is amazing. Thanks for making it!

taskiq/schedule_sources/label_based.py
```diff
@@ -29,6 +29,8 @@ async def get_schedules(self) -> List["ScheduledTask"]:
         schedules = []
         for task_name, task in self.broker.get_all_tasks().items():
             if task.broker != self.broker:
+                # if task broker doesn't match self, something is probably wrong
+                logger.warning(f"Broker for {task_name} ({task.broker}) doesn't match self ({self.broker})")
```
s3rius, line 33:
```suggestion
                logger.warning(
                	f"Broker for {task_name} ({task.broker}) doesn't "
                	f"match scheduler's broker ({self.broker})"
                )
```

taskiq/schedule_sources/label_based.py
```diff
@@ -61,7 +63,10 @@ def post_send(self, scheduled_task: ScheduledTask) -> None:
             return  # it's scheduled task with cron label, do not remove this trigger.
 
         for task_name, task in self.broker.get_all_tasks().items():
-            if task.broker != self.broker or scheduled_task.task_name != task_name:
+            if task.broker != self.broker:
+                # if task broker doesn't match self, something is probably wrong
+                logger.warning(f"Broker for {task_name} ({task.broker}) doesn't match self ({self.broker})")
```
s3rius, line 68:
```suggestion
                logger.warning(
                	f"Broker for {task_name} ({task.broker}) doesn't "
                	f"match scheduler's broker ({self.broker})"
                )
```

s3rius (approved):
These changes should fix the error.

TheTechromancer:
Thanks @s3rius, I added one more check during task registration, which will throw a `TaskRejectedError` if the task is already assigned to a different broker. This should never trigger under normal circumstances.

https://github.com/taskiq-python/taskiq/blob/9f5cf43b2c04ed7abf531f87214157d12e78b05c/taskiq/abc/broker.py#L510-L530

s3rius:
Please run `pre-commit run -a` locally to fix most of style issues.

https://taskiq-python.github.io/contrib.html

TheTechromancer:
There are some weird things going on, where certain linters break the syntax and cause subsequent linters to fail 😬

taskiq/abc/broker.py
```diff
@@ -512,12 +513,19 @@ def _register_task(
...
         But this behaviour can be changed in subclasses.
 
+        This method may raise TaskRejectedError if task has already been
+        registered to a different broker.
+
         :param task_name: Name of a task.
         :param task: Decorated task.
         """
+        if task.broker != self:
+            raise TaskRejectedError(
```
s3rius, line 528: It's not a syntax error. It's [izulu](https://github.com/pyctrl/izulu). We use it for errors. 

Please define separate error class for this error with 

__template__ and broker argument.

TheTechromancer: rgr. This is the syntax error:

image

image

s3rius:
Okay. Actually this PR is fine by me. So, I'm gonna merge it an fix on the main branch.

s3rius (approved):

Issue #427, opened 2025-03-19
Warn people of unset result backend


s3rius:
> @s3rius  Thank you for the quick answer, and it was exactly this. Maybe we should emphasize this in the documentation or put a warning when we use a  `.wait_for` with not Result Backend setup ? 
> 
> I'm closing this anyway, thanks again!  

 _Originally posted by @sorasful in [#426](https://github.com/taskiq-python/taskiq/issues/426#issuecomment-2737643863)_

When the result backend is not specified, `wait_for_result` will return an empty result without actually waiting for the task to be completed. We should generate a warning or raise an exception to inform users that waiting for a result without a result backend is not possible.

igor-gorohovsky:
Can I work on this?

s3rius:
Sure, please do.

igor-gorohovsky:
Thanks. I'll dig into that this week.

PR #430, opened 2025-03-21
Warn if result backend is unset

igor-gorohovsky:
#427

### DIFF
--- a/taskiq/result_backends/dummy.py
+++ b/taskiq/result_backends/dummy.py
@@ -1,13 +1,10 @@
-from logging import getLogger
+import warnings
 from typing import Any, TypeVar
 
 from taskiq.abc.result_backend import AsyncResultBackend
 from taskiq.result import TaskiqResult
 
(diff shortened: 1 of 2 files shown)


igor-gorohovsky:
@s3rius Maybe we should add a warning for other task methods if the result backend is not set. What do you think?

s3rius:
In my opinion the `isinstance` method is quite expensive to call. We can just add `raise in the DummyBackend` itself.

s3rius (changes requested):
In my opinion the isinstance method is quite expensive to call. We can just add raise in the DummyBackend itself.

Because this isinstance call will be triggered for any ResultBackend. Which might decrease performance.

s3rius (approved):

s3rius:
I'm wondering. Maybe it'd be better to use [warn](https://docs.python.org/3/library/warnings.html) module to make warnings appear even if you haven't configured logging.

s3rius (approved):

PR #435, opened 2025-03-25
fix: log configuring when running under spawn and add --log-format

ivolnistov:
**Technical explanation of the issue:**

1. When using 'fork', a copy of the current process is created, including memory and state of all objects. This means logging settings are preserved.
2. When using 'spawn', a new process is created which imports modules fresh. This means all global objects, including loggers, are initialized again.

**Problem in taskiq code:**

1. Logging configuration happens in the parent process, but these settings are not passed to child processes when using 'spawn'.
2. DEBUG logs are not visible because by default loggers in child processes have WARNING or INFO level.

**PR fixes:**

1. This PR fixes the logging issue when using the 'spawn' method in multiprocessing
2. Adds log-format support, which allows more flexible configuration of logger behavior when using spawn

Now DEBUG level logs will be correctly displayed in all child processes, even when using the 'spawn' method, and log formatting settings can be easily customized.

### DIFF
--- a/taskiq/cli/worker/args.py
+++ b/taskiq/cli/worker/args.py
@@ -30,6 +30,7 @@ class WorkerArgs:
     fs_discover: bool = False
     configure_logging: bool = True
     log_level: LogLevel = LogLevel.INFO
+    log_format: str = "[%(asctime)s][%(name)s][%(levelname)-7s][%(processName)s] %(message)s"
     workers: int = 2
     max_threadpool_threads: Optional[int] = None
     max_process_pool_processes: Optional[int] = None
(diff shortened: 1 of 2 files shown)


ivolnistov:
**Technical explanation of the issue:**
1. When using 'fork', a copy of the current process is created, including memory and state of all objects. This means logging settings are preserved.
2. When using 'spawn', a new process is created which imports modules fresh. This means all global objects, including loggers, are initialized again.

**Problem in taskiq code:**
1. Logging configuration happens in the parent process, but these settings are not passed to child processes when using 'spawn'.
2. DEBUG logs are not visible because by default loggers in child processes have WARNING or INFO level.

**PR fixes:**
1. This PR fixes the logging issue when using the 'spawn' method in multiprocessing
2. Adds log-format support, which allows more flexible configuration of logger behavior when using spawn

Now DEBUG level logs will be correctly displayed in all child processes, even when using the 'spawn' method, and log formatting settings can be easily customized.

s3rius (dismissed):

s3rius:
Can you please run `pre-commit run -a`?

s3rius:
Thanks for the pull request. It makes a lot of sense, even though I haven't thought of it.

s3rius (approved):

PR #451, opened 2025-04-19
SmartRetryMiddleware

GefMar:
Problem It Solves

In many real-world scenarios, tasks may fail due to transient issues like network instability or temporary unavailability of external services. Without a retry mechanism, such failures require manual intervention. This middleware provides automatic retry logic, reducing the need for manual restarts and increasing overall system resilience.

Key Features
	•	✅ Retry support with configurable maximum number of attempts (max_retries)
	•	⏱️ Customizable delay between retries (delay)
	•	📈 Support for exponential backoff (use_delay_exponent)
	•	🎲 Optional jitter to randomize delay and reduce retry bursts (use_jitter)
	•	🏷️ Label-based control at the task level:
	•	retry_on_error
	•	max_retries
	•	delay
	•	⏳ Ability to use a custom ScheduleSource for flexible rescheduling
	•	❌ Optionally replaces the task result with NoResultError until retries are exhausted (no_result_on_retry)

How It Works

When a task raises an exception:
	1.	The middleware checks if retries are enabled via the retry_on_error label.
	2.	If the retry count hasn’t reached the maximum, the task is rescheduled with a computed delay.
	3.	If no_result_on_retry is enabled, the task result is temporarily replaced with NoResultError, allowing for deferred handling.


s3rius (approved):

taskiq/middlewares/smart_retry_middleware.py
```diff
@@ -0,0 +1,177 @@
...
+        return delay
+
+    async def on_send(
+        self,
+        kicker: AsyncKicker[Any, Any],
+        message: TaskiqMessage,
+        delay: float,
+    ) -> None:
+        """Execute the task with a delay."""
+        if isinstance(self.schedule_source, ScheduleSource):
```
s3rius, line 106: This condition can be simplified a bit: 

```suggestion
        if self.schedule_source is not None:
```

taskiq/middlewares/smart_retry_middleware.py
```diff
@@ -0,0 +1,177 @@
...
+        :param message: Task message.
+        :param retries: Current retry count.
+        :return: Delay in seconds.
+        """
+        delay = float(message.labels.get("delay", self.default_delay))
+        if self.use_delay_exponent:
+            delay = min(delay * retries, self.max_delay_exponent)
+
+        if self.use_jitter:
+            delay += random.random()  # noqa: S311
```
s3rius, line 95: In order to remove this noqa you can do this:

```suggestion
            delay += secrets.randbelow(1000) / 1000
```

s3rius (dismissed):

s3rius (dismissed):

s3rius (approved):

Issue #318, opened 2024-04-26
Allow configuring hot reload paths


kyboi:
Currently using `--reload` argument will take into account `.gitignore`, but there are files that will trigger a reload even when it isn't necessary. These files must be in the repo so can't be added to the ignore, e.g. docker configuration files.

Perhaps something like uvicorn does, with watchfiles:

https://github.com/encode/uvicorn/blob/master/uvicorn/supervisors/watchfilesreload.py

```
--reload                        Enable auto-reload.
--reload-dir PATH               Set reload directories explicitly, instead
                                  of using the current working directory.
--reload-include TEXT           Set glob patterns to include while watching
                                  for files. Includes '*.py' by default; these
                                  defaults can be overridden with `--reload-
                                  exclude`. This option has no effect unless
                                  watchfiles is installed.
--reload-exclude TEXT           Set glob patterns to exclude while watching
                                  for files. Includes '.*, .py[cod], .sw.*,
                                  ~*' by default; these defaults can be
                                  overridden with `--reload-include`. This
                                  option has no effect unless watchfiles is
                                  installed.
 ```

s3rius:
That might be a good idea. Actually I'd want to implement it the way gunicorn did. Then actually check all files that are actually used by your python project and trigger reload only one of them has changed. Which might be a better fit.

PR #459, opened 2025-05-02
Add --reload-dir argument

montaguegabe:
This is a small change that partially addresses #318, and allows multiple reload dirs to be specified via the CLI instead of using just the CWD, using the `--reload-dir` option, which can be passed multiple times for multiple reload directories.

Why I need this:
I use TaskIQ with tasks loaded from pip packages that I've installed with `pip install -e path/to/local/package` in development, and am constantly editing and having to manually reload TaskIQ due to the fact that the locally-installed package is outside of the CWD.

- If the --reload-dir option is not passed, the default of using the CWD is still used.
- pytest tests still pass
- Committed with pre-commit hooks installed and working
- Adds a logger.debug of the reload dirs being watched.

### DIFF
--- a/taskiq/cli/watcher.py
+++ b/taskiq/cli/watcher.py
@@ -14,12 +14,13 @@ class FileWatcher:  # pragma: no cover
     def __init__(
         self,
         callback: Callable[..., None],
+        path: Path,
         use_gitignore: bool = True,
         **callback_kwargs: Any,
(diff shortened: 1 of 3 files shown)


s3rius (approved):

montaguegabe:
@sebastianjanisch  it is now released.  Thank you very much to s3rius

PR #466, opened 2025-05-20
feat(scheduler): add interval

d3vyce:
Hi!

In this PR, I'm adding the ability to run scheduled tasks every X seconds. This fills a gap that CRON doesn't allow.
To avoid loop invocation spam, I calculate the next run based on which tasks are closest to executing. The next run is therefore dynamic.

This PR is still a WIP, but I welcome feedback to adjust the code if necessary.
I still have to do:
- [ ] Documentation
- [ ] Clean code
- [ ] Tests


d3vyce:
I added `interval` into test_label_discovery, I think it could be a good idea to also test the `interval_next_run` in a complex setup. Like when combining interval tasks with cron and time. Not sure how to do it, but if you have any ideas for implementation, I remain available to integrate them if necessary.
In any case ready for a review and feedback !

PR #467, opened 2025-05-22
Add seconds and milliseconds support in cron_offset

borisalekseev:
Cron is insensitive to seconds. But cron_offset accepts datetime.timedelta, which has seconds in it. This PR adds sensitivity to seconds and milliseconds from cron_offset.

### DIFF
--- a/taskiq/cli/scheduler/run.py
+++ b/taskiq/cli/scheduler/run.py
@@ -85,14 +85,18 @@ def get_task_delay(task: ScheduledTask) -> Optional[int]:
     if task.cron is not None:
         # If user specified cron offset we apply it.
         # If it's timedelta, we simply add the delta to current time.
+        additional_seconds = 0
         if task.cron_offset and isinstance(task.cron_offset, timedelta):
             now += task.cron_offset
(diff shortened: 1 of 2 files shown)


taskiq/cli/scheduler/run.py
```diff
@@ -85,14 +85,18 @@ def get_task_delay(task: ScheduledTask) -> Optional[int]:
     if task.cron is not None:
         # If user specified cron offset we apply it.
         # If it's timedelta, we simply add the delta to current time.
+        additional_seconds = 0
         if task.cron_offset and isinstance(task.cron_offset, timedelta):
             now += task.cron_offset
+            additional_seconds += task.cron_offset.seconds + (
```
s3rius, line 91: Here you use seconds, which might produce unexpected results. 

```suggestion
            additional_seconds += task.cron_offset.total_seconds() + (
```

In this example if you specify minutes and hours it will also correctly work. Or was it implemented like so on purpose?

borisalekseev: There was an even worse mistake here, as you can see. The idea was that additional_seconds is the remainder of the seconds from the entire cron_offset. Now fixed it.

PR #473, opened 2025-06-23
Fix/test warnings

oromogans:
Fixed warnings that appeared when running tests.
For this purpose, I simplified the results module structure a bit.


s3rius (approved):
Seems good to me.

PR #509, opened 2025-10-12
feat: add a tip to the docs about using LabelScheduleSource

spikeninja:


docs/guide/scheduling-tasks.md
```diff
@@ -53,6 +53,12 @@ The scheduler doesn't execute tasks. It only sends them.
 
 You can check list of available schedule sources in the [Available schedule sources](../available-components/schedule-sources.md) section.
 
+::: tip Cool tip!
+
+To use cron functionality, you must include **LabelScheduleSource** even if you use a different source (for example, **ListRedisScheduleSource**).
```
s3rius: Okay, I don't get this tip. 

Did you mean, that in order to specify cron in task labels you have to include LabelScheduleSource?

s3rius: Okay, let's make it a little bit less specific. Because it affects not only crons but schedule labels in general. 

So we can say something like:

```suggestion
To specify schedules (cron or time-based) in the labels of a task definition, you must include **LabelScheduleSource** along with any other sources you use (e.g., **ListRedisScheduleSource**).
```

PR #520, opened 2025-10-22
fix: recursive dict in labels during schedule discovery in LabelScheduleSource

danfimov:
Fixed issues:
- When `labels` are passed inside `schedule`, the `ScheduledTask.labels` field contained a recursive structure  - [discussion](https://t.me/taskiq_py/15255);
- When a user provided `schedule_id` inside `schedule`, it was always replaced with a random UUID.

Also deleted duplicated tests: `tests/schedule_sources/test_label_based.py` are literally the same as in `tests/scheduler/test_label_based_sched.py`.

### DIFF
--- a/docs/available-components/schedule-sources.md
+++ b/docs/available-components/schedule-sources.md
@@ -34,12 +34,13 @@ The format of the schedule label is the following:
 @broker.task(
     schedule=[
         {
-            "cron": "* * * * *", # type: str, either cron or time should be specified.
+            "cron": "*/1 * * * *", # type: str, either cron or time should be specified.
             "cron_offset": None # type: str | timedelta | None, can be omitted.
             "time": None  # type: datetime | None, either cron or time should be specified.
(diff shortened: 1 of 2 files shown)


taskiq/schedule_sources/label_based.py
```diff
@@ -46,8 +46,11 @@ async def startup(self) -> None:
                 if "cron" not in schedule and "time" not in schedule:
                     continue
                 labels = schedule.get("labels", {})
-                labels.update(task.labels)
-                schedule_id = uuid.uuid4().hex
+
+                task_labels = {k: v for k, v in task.labels.items() if k != "schedule"}
+
+                labels.update(task_labels)
+                schedule_id = schedule.get("schedule_id", uuid.uuid4().hex)
```
s3rius: Can you add to the doc, that now you can specify schedule_id?

s3rius (dismissed):

PR #525, opened 2025-10-26
feat: Add OpenTelemetry instrumentation

soapun:


docs/README.md
```diff
@@ -14,13 +14,13 @@ actions:
...
     - name: "google-site-verification"
-      content: "hQCR5w2tmeuOvYIYXsOYU3u4kLNwT86lnqltANYlRQ0"
+      content: "hQCR5w2tmeuOvYIYXsOYU3u4kLNwT86lnqltANYlRQ0" # pragma: allowlist secret
   - - meta
     - name: "msvalidate.01"
-      content: "97DC185FE0A2F5B123861F0790FDFB26"
+      content: "97DC185FE0A2F5B123861F0790FDFB26" # pragma: allowlist secret
   - - meta
     - name: "yandex-verification"
-      content: "9b105f7c58cbc920"
```
danfimov: Why `pragma: allowlist secret` is here?

As I see, we already exclude this file from `detect-secrets` pre-commit hook check. Is some other linter checks fails on this lines?

taskiq/package.py
```diff
@@ -0,0 +1,2 @@
+# for compatibility with opentelemetry-instrumentation
+_instruments = ("taskiq >= 0.11.19",)
```
danfimov: Just wondering: do we need to update version in this file on every release?

docs/README.md
danfimov: In any case - it's not a blocker to merge this MR. I will deal with it later if we don't find the root cause of this strange pre-commit hook behaviour.

taskiq/instrumentation.py
```diff
@@ -0,0 +1,138 @@
...
+)
+from opentelemetry.instrumentation.utils import unwrap
+from opentelemetry.metrics import MeterProvider
+from opentelemetry.trace import TracerProvider
+from wrapt import wrap_function_wrapper
+
+from taskiq import AsyncBroker
+from taskiq.middlewares.opentelemetry_middleware import OpenTelemetryMiddleware
+
+if TYPE_CHECKING:
```
danfimov: Why do we have an empty `if` statement here?

docs/README.md
soapun: I guess the root cause is me commiting from windows :)
On mac the exclusions are working as expected

taskiq/middlewares/opentelemetry_middleware.py
```diff
@@ -0,0 +1,299 @@
...
+_TASK_RETRY_REASON_KEY = "taskiq.retry.reason"
+_TASK_NAME_KEY = "taskiq.task_name"
+
+
+def set_attributes_from_context(span: Span, context: Dict[str, Any]) -> None:
+    """Helper to extract meta values from a Taskiq Context."""
+    if not span.is_recording():
+        return
+
+    for key in TASKIQ_CONTEXT_ATTRIBUTES:
```
s3rius: Why do we track only particular labels? Let's just iterate over all available labels. 

Also, I think that we also want to extract args and kwargs to set it on the span. For better tracing.  Instead of passing labels, you can pass TaskiqMessage itself and get everthing from it.

taskiq/middlewares/opentelemetry_middleware.py
```diff
@@ -0,0 +1,299 @@
...
+        operation_name = f"{_TASK_SEND}/{message.task_name}"
+        span = self._tracer.start_span(operation_name, kind=trace.SpanKind.PRODUCER)
+
+        if span.is_recording():
+            span.set_attribute(_TASK_TAG_KEY, _TASK_SEND)
+            span.set_attribute(SpanAttributes.MESSAGING_MESSAGE_ID, message.task_id)
+            span.set_attribute(_TASK_NAME_KEY, message.task_name)
+            set_attributes_from_context(span, message.labels)
+
+        activation = trace.use_span(span, end_on_exit=True)
```
s3rius: Here's a small quetsion about spans. It seems like you use spans as events. 

From my understanding it makes sense to only have 1 single span for task execution. And inside of that span we can use events to show what was going on during the exeuction. 

I'm not sure if it's the best way. I need to try it out.

s3rius (changes requested):

taskiq/middlewares/opentelemetry_middleware.py
s3rius: 🤔 🤔 🤔 

Let's leave it as it is right now. We might reconsider it later.

s3rius (changes requested):

PR #568, opened 2025-12-18
Fixed reload-loop.

s3rius:
I wanted to make filewatching more precise to reduce number of files we are watching. But using modules for that is not an option. Unless we implement restart functionality in processes themselves. Which is much harder and seems like an overengineered solutions. 

So instead I just added more filters to avoid reload loops.


epicwhale:
This fix does solve the reload issues ive been facing in dev! Could this be tagged as a release please? :)

PR #600, opened 2026-03-03
fix: check readiness before timeout

ans-artal:
I think that when a task completes as the timeout expire, the code should return the result rather than raise.


s3rius:
Yeah, that could help a bit. Thanks for your contribution.

s3rius (approved):

PR #606, opened 2026-04-01
Get rid of izulu

mahenzon:
idk why we need izulu here. It's a strange lib.

What concerns me:
- why we froze version 0.50.0 when there's already 0.75.0?
- why all latest releases of izulu are broken?
- why rely on a 3rd-party library that has exactly one maintainer?
- why [izulu claims](https://github.com/pyctrl/izulu/blob/f51c8f1f420094f2122a2e4ead2ac70f6bc0c5b6/pyproject.toml#L12) that it's MIT-licensed, but in the [license file](https://github.com/pyctrl/izulu/blob/main/LICENSE) it has custom license?

taskiq doesn't even utilize so much from the izulu library.

Let's get rid of it. It's just about 70 lines of code.

Here's my MIT-licensed source code that can be just copy-pasted into the project. Remember the [left-pad](https://habr.com/ru/articles/280099/).

https://gist.github.com/mahenzon/5ac32267ebd6d29a44dd52bb94a9630c

### DIFF
--- a/taskiq/error.py
+++ b/taskiq/error.py
@@ -1,8 +1,20 @@
 """Minimal exception templating used by taskiq exceptions."""
 
+import sys
 from string import Formatter
 
+if sys.version_info >= (3, 11):
+    from typing import dataclass_transform
+else:
+    from typing_extensions import dataclass_transform
 
+
+@dataclass_transform(
+    eq_default=False,
+    order_default=False,
+    kw_only_default=True,
+    frozen_default=False,
+)
 class Error(Exception):
     """Base templated exception compatible with taskiq needs."""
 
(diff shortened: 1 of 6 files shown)


taskiq/error.py
```diff
@@ -0,0 +1,70 @@
+"""Minimal exception templating used by taskiq exceptions."""
+
+from string import Formatter
+
+
+class Error(Exception):
```
s3rius, line 18: Make it a dataclass_transform, so we could see it's attributes needed to be set. 

https://peps.python.org/pep-0681/

It will add this thing a proper mypy validation.

mahenzon: Thank you. Made it look more like [izulu's implementation](https://github.com/pyctrl/izulu/blob/1efeb46eb4de0848f536ab8f1e1fa7cf3bced1c0/izulu/root.py#L87-L94)

s3rius (approved):

Issue #602, opened 2026-03-22
Feature Request: Add NNG (Nanomsg Next Gen) Broker Support


alexted:
### Is your feature request related to a problem?

Currently, `taskiq` provides support for a ZeroMQ broker. However, the existing ZMQ implementation comes with certain architectural limitations, specifically regarding seamless asynchronous integration, dynamic worker discovery, and the complexity of managing socket lifecycles when nodes randomly disconnect or scale.

For high-performance, brokerless distributed systems, developers often need a lighter, more modern alternative to handle messaging topologies without the boilerplate and fragility of raw ZMQ socket management inside `asyncio` loops.

### Describe the solution you'd like

I would like to propose adding [NNG](https://github.com/nanomsg/nng) ([Nanomsg Next Generation](https://nng.nanomsg.org/)) broker support to `taskiq`, potentially utilizing Python bindings like `pynng`. 

NNG is a modern, lightweight rewrite of Nanomsg that addresses many of the scalability and state-handling issues found in older ZMQ architectures. A dedicated NNG broker would provide:
* Clean, native `asyncio` integration out-of-the-box.
* Decoupled dialers and listeners, meaning far more robust reconnection patterns when workers crash or restart.
* A clean, lightweight environment for fast inter-process and network communications, ideal for modern microservice architectures.

### Describe alternatives you've considered

1. **Using the existing ZeroMQ broker:** This currently feels restrictive due to the boilerplate required to safely manage central proxy topologies and the lack of robust, automated reconnect handling in complex asynchronous contexts.
2. **Writing a custom broker from scratch:** While possible, NNG is a standard, highly efficient protocol. Having official or community-supported integration in the `taskiq` ecosystem would greatly benefit developers building fast, lightweight asynchronous services.

### Additional context
Adding an NNG backend (e.g., `taskiq-nng`) would be a huge win for developers looking for reliability and performance in their message queues without deploying heavy infrastructure like Redis, Kafka or NATS, keeping the technology stack lean and efficient.

### Which component would this affect?

Broker

s3rius:
Hello, @alexted and thanks for your interest in the project. I actually appreciate that you showed me this broker. I've never heard of it, and after trying I should admit, that I enjoyed how easy it was to integrate and how cool the technology is. 

[...]

PR #607, opened 2026-04-11
Added NNG broker.

s3rius:
Ref: #602.

### DIFF
--- a/taskiq/brokers/nng_broker.py
+++ b/taskiq/brokers/nng_broker.py
@@ -1,0 +1,48 @@
+from collections.abc import AsyncGenerator
+
+import pynng
+
+from taskiq.abc.broker import AsyncBroker
+from taskiq.message import BrokerMessage
+
+
+class NNGBroker(AsyncBroker):
+    """
+    NanoMSG next generation broker.
+
+    This broker is very much alike to the ZMQ broker,
+    It has a similar Idea, but slightly different
+    implementation.
+    """
+
+    def __init__(self, addr: str) -> None:
+        """
+        Initialize the broker.
+
+        :param addr: address which is used by both worker and client.
+        """
+        super().__init__()
+        self.socket = pynng.Pair1(polyamorous=True)
+        self.addr = addr
+
+    async def startup(self) -> None:
+        """Start the socket."""
+        await super().startup()
+        if self.is_worker_process:
+            self.socket.listen(self.addr)
+        else:
+            self.socket.dial(self.addr, block=True)
+
+    async def shutdown(self) -> None:
+        """Close the socket."""
+        await super().shutdown()
+        self.socket.close()
+
+    async def kick(self, message: BrokerMessage) -> None:
+        """Send a message."""
+        await self.socket.ascend(message.message)
+
+    async def listen(self) -> AsyncGenerator[bytes, None]:
+        """Infinite loop that receives messages."""
+        while True:
+            yield await self.socket.arecv()
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.14' 
+          python-version: '3.14'
       - id: setup-uv
         name: Setup UV
         uses: astral-sh/setup-uv@v7
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v6
         with:
-          python-version: '${{ matrix.py_version }}' 
+          python-version: '${{ matrix.py_version }}'
       - id: setup-uv
         name: Setup UV
         uses: astral-sh/setup-uv@v7
--- a/docs/available-components/brokers.md
+++ b/docs/available-components/brokers.md
@@ -43,6 +43,36 @@ pip install "taskiq[zmq]"
 
 :::
 
+## NNG broker
+
+NNG broker is very much alike to the ZMQ broker. It is as lightweight as ZMQ, but uses NNG as an underlying socket
+management solution.
+
+For more information visit their website at [nng.nanomsg.org](https://nng.nanomsg.org/).
+
+::: caution Be careful!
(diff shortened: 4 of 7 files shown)


taskiq/brokers/nng_broker.py
```diff
@@ -0,0 +1,48 @@
...
+            self.socket.dial(self.addr, block=True)
+
+    async def shutdown(self) -> None:
+        """Close the socket."""
+        await super().shutdown()
+        self.socket.close()
+
+    async def kick(self, message: BrokerMessage) -> None:
+        """Send a message."""
+        await self.socket.ascend(message.message)
```
alexted, line 43: The correct `pynng` method for async sending is `asend()`. Currently, this will raise an `AttributeError` on the very first task dispatch.

alexted (changes requested):
The new `NNGBroker` is not exported in `taskiq/__init__.py` and `taskiq/brokers/__init__.py`. We need to export it to keep the public API consistent with `ZeroMQBroker` and others, so users can just do `from taskiq import NNGBroker`.

alexted (changes requested):
Since this is a core networking component, we strictly need at least basic integration tests (e.g., in `tests/brokers/`) to verify `startup`, `kick`, `listen`, and `shutdown` behaviors.

taskiq/brokers/nng_broker.py
```diff
@@ -0,0 +1,48 @@
...
+        self.socket = pynng.Pair1(polyamorous=True)
+        self.addr = addr
+
+    async def startup(self) -> None:
+        """Start the socket."""
+        await super().startup()
+        if self.is_worker_process:
+            self.socket.listen(self.addr)
+        else:
+            self.socket.dial(self.addr, block=True)
```
alexted, line 34: **NNG** (like most brokers) can be susceptible to _slow joiner syndrome_ during `dial` and `listen`. If we `kick` a task immediately after `dial`, the message might be dropped or cause an error before the connection is fully established. To make the initial transmission more robust, should we consider adding a brief timeout or a socket state check?
Just thinking out loud - pure theory.

taskiq/brokers/nng_broker.py
```diff
@@ -0,0 +1,48 @@
...
+class NNGBroker(AsyncBroker):
+    """
+    NanoMSG next generation broker.
+
+    This broker is very much alike to the ZMQ broker,
+    It has a similar Idea, but slightly different
+    implementation.
+    """
+
+    def __init__(self, addr: str) -> None:
```
alexted, line 18: The `addr` parameter is accepted but remains unvalidated. Given that **Taskiq** relies heavily on **Pydantic**, it would be fitting to add basic validation for the connection string (e.g., checking for `ipc://` or `tcp://` prefixes).

taskiq/brokers/nng_broker.py
```diff
@@ -0,0 +1,48 @@
+from collections.abc import AsyncGenerator
+
+import pynng
```
alexted, line 3: The dependency is added as an **extra** in `pyproject.toml`, but the code is missing a guard import. 
```python
try:
    import pynng
except ImportError:
    raise ImportError("Install 'taskiq[nng]' to use NNGBroker.")
```
Without it the user will get a confusing `ModuleNotFoundError` only when they try to initialize the class.

alexted (changes requested):

taskiq/brokers/nng_broker.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+
+    def __init__(self, addr: str) -> None:
+        """
+        Initialize the broker.
+
+        :param addr: address which is used by both worker and client.
+        """
+        super().__init__()
+        self.socket = pynng.Pair1(polyamorous=True)
```
alexted, line 25: There is a fundamental architectural bottleneck that needs to be addressed to make this broker production-ready.

The current use of `pynng.Pair1` with `listen()` inside the worker process forces the `-w 1` limitation. In a distributed system, if multiple worker processes try to `bind` (listen) to the same port, it leads to address conflicts or, worse, duplicated task execution (where every worker picks up the same message). This is exactly why the existing ZeroMQ broker in `taskiq` is limited to a single worker.

The Solution: **Router/Dealer + Proxy Device**
To unlock horizontal scaling (`-w 100`) and bi-directional feedback (metrics, statuses), we should move to a **centralized Proxy topology** using **Router/Dealer** sockets.

- **The Proxy (Device)**: A lightweight mediator that `listen()`s on two ports (Frontend for clients, Backend for workers). It uses `pynng.device()` to pump messages between them with native NNG Round-Robin load balancing.

- **The Clients & Workers**: Both become purely outbound - they only `dial()` the proxy. This eliminates port conflicts between worker processes.

**Proposed Implementation**
The broker should automatically attempt to start the proxy if it's running locally, or simply connect to it.

Broker Core Logic:
```python
import pynng
import asyncio
import threading

class NNGBroker(AsyncBroker):
    async def startup(self) -> None:
        # 1. Try to start the internal proxy device 
        # (Only the first process succeeds, others will fail silently if port is bound)
        self._start_proxy_if_needed()

        if self.is_worker_process:
            # Workers connect to the backend
            self.socket = pynng.Dealer0(dial=self.worker_url)
        else:
            # Clients connect to the frontend
            self.socket = pynng.Dealer0(dial=self.client_url)

    def _start_proxy_if_needed(self):
        try:
            # Simple check/start logic for pynng.device
            # This allows horizontal scaling without manual proxy management
            def run_device():
                with pynng.Router0(listen=self.client_url) as front, \
                     pynng.Dealer0(listen=self.worker_url) as back:
                    pynng.device(front, back)
            
            threading.Thread(target=run_device, daemon=True).start()
        except pynng.AddressInUse:
            pass # Proxy is already running
```
Worker Feedback:
```python
async def listen(self):
    async for message in self.socket:
        # Asynchronous feedback is now possible!
        await self.socket.asend(b"Status: Processing") 
        yield message
```
By adopting the **Router/Dealer + Proxy** pattern, we overcome the limitations of the ZeroMQ implementation and provide a truly scalable NNG transport for the ecosystem.

UPD: I have prepared an implementation based on my understanding of how this could and should work. You can find it here - https://github.com/taskiq-python/taskiq/pull/612.

alexted (changes requested):

PR #612, opened 2026-04-26
Feat: Improved implementation of NNG broker support

alexted:
Refactoring the https://github.com/taskiq-python/taskiq/pull/607


taskiq/brokers/nng/hub.py
```diff
@@ -0,0 +1,463 @@
...
+                    logger.info("Requeued %d tasks from dead workers", recovered)
+            except asyncio.CancelledError:
+                raise
+            except Exception:
+                logger.exception("Reaper loop error")
+
+
+# ── standalone CLI entry point ────────────────────────────────────────────────
+
+def _build_config() -> HubConfig:
```
s3rius: Instead of defining a stand-alone file to run, let's make it a sub-command of the taskiq itself. 

You can do it by following our CLI extending guide. 

https://taskiq-python.github.io/extending-taskiq/cli.html

taskiq/brokers/nng/hub.py
```diff
@@ -0,0 +1,463 @@
...
+    independent ``nng_ctx`` contexts running concurrently.  Each context
+    handles one request-reply at a time, so N workers can
+    register/heartbeat/ack simultaneously without queuing behind each other.
+
+    **Data plane** — One ``Push0`` socket per registered worker, dialed to
+    the worker's own ``Pull0`` listen address.  The hub explicitly targets
+    the least-loaded worker instead of relying on NNG round-robin.
+
+    **State** — :class:`~taskiq.brokers.nng.storage.InMemoryStore`.  All
+    store operations are synchronous and execute directly on the asyncio event
```
s3rius, line 90:
```suggestion
    store operations are asynchronous and execute directly on the asyncio event
```

taskiq/brokers/nng/broker.py
```diff
@@ -0,0 +1,328 @@
...
+        if self._ctrl_sock is None:
+            raise RuntimeError("Broker is not started")
+        payload: dict[str, Any] = {
+            "task_id": message.task_id,
+            "task_name": message.task_name,
+            "payload_b64": base64.b64encode(message.message).decode("ascii"),
+            "labels": message.labels,
+            "lease_id": "",  # hub assigns the real lease_id at dispatch time
+            "attempts": int(message.labels.get("attempts", 0)),
+            "max_retries": int(
```
s3rius: Since we have retry-middlewares that do exactly that, I guess we can remove retry functionality from the broker and the hub.

taskiq/brokers/nng/broker.py
```diff
@@ -0,0 +1,328 @@
...
+            except Exception as exc:
+                logger.error("Malformed task envelope discarded: %s", exc)
+                continue
+
+            task_id = envelope.task_id
+            worker_id = self.worker_id
+            lease_id = envelope.lease_id  # hub-assigned; correct by construction
+
+            async def _ack(
+                _task_id: str = task_id,
```
s3rius, line 288: Instead of creating those parameters as defaults, just use them inside the function. In that case your closure will borrow variables from the outer environment. It will make it more rock solid to use.

taskiq/brokers/nng/protocol.py
```diff
@@ -0,0 +1,159 @@
...
+    Workers must echo it back in the ACK so the hub can validate
+    that the ack is not stale (e.g. after lease expiry and requeue).
+    """
+
+    task_id: str
+    task_name: str
+    payload_b64: str
+    labels: dict[str, Any] = field(default_factory=dict)
+    lease_id: str = ""
+    attempts: int = 0
```
s3rius: Retry-related again.

taskiq/brokers/nng/storage.py
```diff
@@ -0,0 +1,666 @@
...
+    capacity: int
+    inflight: int = 0
+    last_seen: float = 0.0
+    heartbeat_interval: float = 5.0
+    lease_timeout: float = 15.0
+    draining: bool = False
+    status: str = "starting"
+    version: str = "unknown"
+
+    def as_dict(self) -> dict[str, Any]:
```
s3rius, line 77: You don't need to define a function for that. There's a built-in [asdict](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) function for dataclasses.

taskiq/brokers/nng/protocol.py
```diff
@@ -0,0 +1,159 @@
...
+            payload=data.get("payload", {}),
+            error=data.get("error"),
+        )
+
+    def to_bytes(self) -> bytes:
+        """Serialise to JSON bytes."""
+        return json.dumps(
+            {"ok": self.ok, "payload": self.payload, "error": self.error},
+            separators=(",", ":"),
+            ensure_ascii=False,
```
s3rius, line 137: Separators will save you like few bytes. But it might make messages more readable. 
 
```suggestion
            ensure_ascii=False,
```

taskiq/brokers/nng/broker.py
```diff
@@ -0,0 +1,328 @@
...
+        :param message: broker message to submit.
+        :raises RuntimeError: if the broker has not been started or the hub
+            rejects the submission (e.g. queue full).
+        """
+        if self._ctrl_sock is None:
+            raise RuntimeError("Broker is not started")
+        payload: dict[str, Any] = {
+            "task_id": message.task_id,
+            "task_name": message.task_name,
+            "payload_b64": base64.b64encode(message.message).decode("ascii"),
```
s3rius, line 240:
```suggestion
            "payload_b64": base64.b64encode(message.message).decode("utf-8"),
```

UTF-8 is used in all other places. Let's keep it everywhere to not have any encoding\decoding issues.

taskiq/brokers/nng/hub.py
```diff
@@ -0,0 +1,463 @@
...
+        self._worker_push.clear()
+        if self._ctrl_sock is not None:
+            with suppress(Exception):
+                self._ctrl_sock.close()
+        logger.info("NNG hub stopped")
+
+    # ── control plane ─────────────────────────────────────────────────────────
+
+    async def _control_handler(self, ctx: Any) -> None:
+        """Run one Rep0 context: receive → dispatch → reply, in a loop."""
```
s3rius:
```suggestion
        """Run one Rep0 context: receive -> dispatch -> reply, in a loop."""
```

s3rius (changes requested):
That's a whole broker-server implementation, which I'm not quite sure if we want to have it in the main repo. 

I thought that it would be like ZeroMQ, but it's much more impressive. Maybe we might consider moving it to another package to keep the main lib thin.

alexted:
> That's a whole broker-server implementation, which I'm not quite sure if we want to have it in the main repo.
> 
> I thought that it would be like ZeroMQ, but it's much more impressive. Maybe we might consider moving it to another package to keep the main lib thin.

I intended this PR as a full implementation proposal, rather than a requirement to merge it as-is. Feel free to adopt it fully, partially, or not at all, depending on what aligns best with the project’s direction.

taskiq/brokers/nng/broker.py
alexted: **Base64** output is **ASCII** by definition, `.decode("ascii")` is stricter and semantically more accurate. "Consistency" is not relevant here.

taskiq/brokers/nng/protocol.py
alexted: For the wire format, compactness is the norm, not a micro-optimization. These are bytes on the socket, not logs.

taskiq/brokers/nng/broker.py
alexted: Default arguments here are the canonical fix for **late binding** in Python. `task_id/lease_id/worker_id` are reassigned on every iteration, and without snapshotting via defaults, a delayed `ack` will capture someone else’s values.

taskiq/brokers/nng/hub.py
alexted: The methods are actually `def`, not `async def`, so the wording is correct.

alexted:
> That's a whole broker-server implementation, which I'm not quite sure if we want to have it in the main repo.
> 
> I thought that it would be like ZeroMQ, but it's much more impressive. Maybe we might consider moving it to another package to keep the main lib thin.

Strategically valid, you are right.

taskiq/brokers/nng/broker.py
alexted: removed

taskiq/brokers/nng/protocol.py
alexted: removed

taskiq/brokers/nng/storage.py
alexted: reworked

taskiq/brokers/nng/hub.py
alexted: moved

taskiq/brokers/nng/hub.py
alexted: replaced

PR #620, opened 2026-05-07
Fixed prometheus multiprocessing registry

FadingFog:
Replaces global prometheus REGISTRY with a new CollectorRegistry

### DIFF
--- a/taskiq/middlewares/prometheus_middleware.py
+++ b/taskiq/middlewares/prometheus_middleware.py
@@ -84,7 +84,10 @@ class PrometheusMiddleware(TaskiqMiddleware):
         This function starts prometheus server.
         It starts it only in case if it's a worker process.
         """
-        from prometheus_client import CollectorRegistry, start_http_server  # noqa: PLC0415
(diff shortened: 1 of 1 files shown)


s3rius:
First off, can you please run `uv run pre-commit run -a` to fix errors. 

And, second  thing. Can you please tell me how to run it properly? Because right now with the following code:

```python
from taskiq_redis import ListQueueBroker

from taskiq.middlewares import PrometheusMiddleware

broker = ListQueueBroker("redis://localhost").with_middlewares(PrometheusMiddleware())
```

If you run it with 

`taskiq worker a:broker`

Then `localhost:9000` returns empty page.

FadingFog:
> First off, can you please run `uv run pre-commit run -a` to fix errors.

Fixed
> Then `localhost:9000` returns empty page.

It's happening because we create a new registry that doesn't have default python metrics, so there're no metrics at the time of worker startup. You can execute a task and then the metrics will be populated

s3rius:
Yeah, I tested it out, looks good.

s3rius (approved):

PR #624, opened 2026-05-16
experiment separate Broker + Router

GefMar:


taskiq/flow.py
```diff
@@ -0,0 +1,58 @@
...
+    name: str
+    kind: FlowKind = FlowKind.QUEUE
+    options: dict[str, Any] = field(
+        default_factory=dict,
+        compare=False,
+        hash=False,
+    )
+
+    @classmethod
+    def queue(cls, name: str, **options: Any) -> "Flow":
```
s3rius: I don't get what are odds in using different flow kinds? How are they going to differ?

docs/examples/router/multiple_brokers.py
```diff
@@ -0,0 +1,65 @@
...
+
+@default_broker.task(task_name="examples.send_email", domain="notifications")
+async def send_email(user_id: int, template: str) -> str:
+    """Pretend to render and send an email."""
+    return f"{template} email sent to user {user_id}"
+
+
+router.route_task(
+    send_email.task_name,
+    broker="priority",
```
s3rius: Referencing by a magic string is not a good idea. I guess placing a broker value in here would be much better.

docs/examples/router/multiple_brokers.py
```diff
@@ -0,0 +1,65 @@
...
+)
+
+
+async def _main() -> None:
+    await default_broker.startup()
+    await priority_broker.startup()
+    try:
+        direct_result = await send_email(7, "welcome")
+
+        routed_task = await send_email.kiq(7, "welcome")
```
s3rius: It says that the task is routed, but it's very implicit.

docs/examples/router/multiple_brokers.py
```diff
@@ -0,0 +1,65 @@
...
+        routed_result = await routed_task.wait_result(timeout=2)
+
+        bulk_task = await send_email.kicker().with_route(
+            "default",
```
s3rius: String references.

s3rius (changes requested):
Well. Although the intention is good for sure. 

But I see several problems with the solution you provided. Most importantly I see that brokers are referenced by strings almost everywhere. Which is problematic.

Also, since everything is implicit, it's hard to resolve conflicts in this setup. Like, what if I have multiple routers. I guess by design it's highly advised to only use one instance of router per project, but most probably people would not care about this constraint, since it's not enforced.

I'm not  quite sure about Flow concept. Although multiple queue support is highly anticipated by the community, and I'm fully up for its support. But this implementation lacks some fundamental abilities for future expansion. Different brokers might have their own different parameters on Queues. For example RabbitMQ has lots of things that it can do with a queue. 

And I'm not quite sure about flow_kind. It seems a bit redundant. Or I didn't catch the meaning of it.

I still like the intention. But I guess it need to be redesigned.

PR #625, opened 2026-05-18
fix: round-trip timedelta cron offset (#605)

### LINKED ISSUE #605: Scheduler's CronSpec does not work with datetime.timezone offset
### Taskiq version

0.12.1

### Python version

Python 3.13

### OS

macOS

### What happened?

When serializing and subsequently validating `CronSpec` with offset as `datetime.timezone`, the validated instance contains string representation of the offset, which is not valid:

```
>>> CronSpec.model_validate(CronSpec(offset=datetime.datetime.now().astimezone().tzinfo.utcoffset(None)).model_dump(mode="json"))
CronSpec(minutes='*', hours='*', days='*', months='*', weekdays='*', offset='PT2H')
>>> datetime.datetime.now().astimezone().tzinfo.utcoffset(None)
datetime.timedelta(seconds=7200)
```

Kicking the tasks uses `.model_dump(mode="json")`, which is what breaks the timezone-typed value.

### Relevant log output

```shell

```

```python

```

sfrangulov:
## What

A `timedelta` cron offset breaks after a serialize→deserialize round-trip. pydantic serializes a `timedelta` to an ISO-8601 duration string (e.g. `"PT4H"`); because the offset fields are typed `str | timedelta`, on reload the value stays a `str`, so `is_cron_task_now` takes the timezone-name branch and calls `ZoneInfo("PT4H")` → the schedule breaks.

Reproduction (from #605):

```python
>>> CronSpec.model_validate(CronSpec(offset=datetime.now().astimezone().tzinfo.utcoffset(None)).model_dump(mode="json"))
CronSpec(..., offset='PT2H')   # expected: timedelta(seconds=7200)
```

## Fix

Add a shared `parse_cron_offset` validator (`scheduler/scheduled_task/validators.py`) that restores a `timedelta` from its serialized ISO-8601 duration form, while leaving genuine timezone names (e.g. `"US/Eastern"`) and all other values untouched — a timezone name is not a valid duration, so it is safely passed through.

Applied as a before-validator on:

- `CronSpec.offset`
- `ScheduledTask.cron_offset` (pydantic v1 and v2 models)

## Tests

`tests/scheduler/test_cron_offset_roundtrip.py` — round-trip for `CronSpec` and `ScheduledTask` with `timedelta` offsets, plus regression guards that timezone-name offsets are preserved.

## Verification (local)

- New + existing scheduler/cron tests pass (`tests/scheduler/`, `tests/cli/scheduler/`).
- `ruff`, `black`, `mypy` clean on changed files.
- Verified on pydantic v2 (environment). The pydantic v1 branch mirrors the existing dual-version validator idiom already used in `taskiq/result/result.py` and shares the same helper.

Closes #605

### DIFF
--- a/taskiq/scheduler/scheduled_task/cron_spec.py
+++ b/taskiq/scheduler/scheduled_task/cron_spec.py
@@ -1,6 +1,21 @@
+from collections.abc import Callable
 from datetime import timedelta
+from typing import Any
 
 from pydantic import BaseModel
+
+from taskiq.compat import IS_PYDANTIC2
+from taskiq.scheduler.scheduled_task.validators import parse_cron_offset
+
+_offset_before_validator: Callable[..., Any]
+if IS_PYDANTIC2:
+    from pydantic import field_validator
+
+    _offset_before_validator = field_validator("offset", mode="before")
+else:
+    from pydantic import validator
+
+    _offset_before_validator = validator("offset", pre=True)
 
 
 class CronSpec(BaseModel):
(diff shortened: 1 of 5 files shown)


danfimov:
We took a slightly different approach and solved the problem using a discriminator here: https://github.com/taskiq-python/taskiq/pull/679

PR #629, opened 2026-05-27
Fix InMemoryBroker to use startup and shutdown middleware hooks

PythonFZ:
A copy/paste approach + test similar to https://github.com/taskiq-python/taskiq/pull/358

Instead of relying on `super` calls this PR just copy + pastes

https://github.com/taskiq-python/taskiq/blob/9f8db96bf52947939651b4a8f0b2fdc663c6ca33/taskiq/abc/broker.py#L196-L198

and 

https://github.com/taskiq-python/taskiq/blob/9f8db96bf52947939651b4a8f0b2fdc663c6ca33/taskiq/abc/broker.py#L217-L219

into the methods of `class InMemoryBroker(AsyncBroker)` and adds a simple test case checking ensuring `startup` and `shutdown` work and the order of execution is correct.

### DIFF
--- a/taskiq/brokers/inmemory_broker.py
+++ b/taskiq/brokers/inmemory_broker.py
@@ -5,6 +5,7 @@ from concurrent.futures import ThreadPoolExecutor
 from typing import Any, TypeVar
 
 from taskiq.abc.broker import AsyncBroker
+from taskiq.abc.middleware import TaskiqMiddleware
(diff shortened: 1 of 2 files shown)


s3rius (approved):

PythonFZ:
@s3rius Thanks for approving the PR. Could you re-run the CI / merge `main` again. The error doesn't seem to be related to the changed I made.

danfimov (approved):

PR #630, opened 2026-05-28
fix: `prefetcher` leaks one message at `max_tasks_to_execute` boundary

### LINKED ISSUE #528: max-prefetch argument doesn't work as expected
Thank you for this amazing project.

The max-prefetch argument doesn't work as expected. If there are tasks to be consumed, the prefetch queue will be bigger than the argument max-prefetch by 1.

Look at this part of the code in the `receiver.prefetcher` [here](https://github.com/taskiq-python/taskiq/blob/9339f9d50e436406acb28caa5625098943bfadcc/taskiq/receiver/receiver.py#L367); and consider the max-prefetch to be 0.

```python
        while True:
            if finish_event.is_set():
                break
            try:
                await self.sem_prefetch.acquire()
                if (
                    self.max_tasks_to_execute
                    and fetched_tasks >= self.max_tasks_to_execute
                ):
                    logger.info("Max number of tasks executed.")
                    break
                # Here we wait for the message to be fetched,
                # but we make it with timeout so it can be interrupted
                done, _ = await asyncio.wait({current_message}, timeout=0.3)
                # If the message is not fetched, we release the semaphore
                # and continue the loop. So it will check if finished event was set.
                if not done:
                    self.sem_prefetch.release()
                    continue
                # We're done, so now we need to check
                # whether task has returned an error.
                message = current_message.result()
                ## << consider the case where we already consumed a message from the previous iteration
                ## << and so message is an Ackable instance
                current_message = asyncio.create_task(iterator.__anext__())  
                ## << when the interpreter arrives here, the task to consume another message is created but not awaited
                ## << so we don't have control over when it will start consuming via the broker.
                ## << usually the broker will have enough time to consume a request and it will yield its value after the next await (line done, _ = await asyncio.wait({current_message}, timeout=0.3)
                fetched_tasks += 1
                await queue.put(message)
            except (asyncio.CancelledError, StopAsyncIteration):
                break
```

I think we shouldn't create a task an wait for it separately like we currently do.
we should something like:

```python
        while True:
            if finish_event.is_set():
                break
            try:
                await self.sem_prefetch.acquire()
                if (
                    self.max_tasks_to_execute
                    and fetched_tasks >= self.max_tasks_to_execute
                ):
                    logger.info("Max number of tasks executed.")
                    break
                # Here we wait for the message to be fetched,
                # but we make it with timeout so it can be interrupted
                ## << create current task and consume it right away
                current_message = asyncio.create_task(iterator.__anext__())  # type: ignore
                done, _ = await asyncio.wait({current_message}, timeout=0.3)
                # If the message is not fetched, we release the semaphore
                # and continue the loop. So it will check if finished event was set.
                if not done:
                    self.sem_prefetch.release()
                    continue
                # We're done, so now we need to check
                # whether task has returned an error.
                message = current_message.result()
                
                fetched_tasks += 1
                await queue.put(message)
            except (asyncio.CancelledError, StopAsyncIteration):
                break
```

I can do a PR and do a proper fix if you agree

PythonFZ:
Fix using `max_tasks_to_execute` causes the `prefetcher` to pop tasks from the queue it has no intention of processing.

Closes #528.

### DIFF
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -436,9 +436,14 @@ class Receiver:
                     break
                 except Exception:
                     logger.exception("Error while prefetching.")
-                    current_message = None
-                    iterator = self.broker.listen()
-                    self.sem_prefetch.release()
+                    # current_message set => fetch failed before enqueue, so we
+                    # still own the permit and a (possibly broken) iterator.
+                    # Otherwise it's queued and the runner owns the permit;
+                    # releasing here would leak a prefetch slot.
+                    if current_message is not None:
+                        current_message = None
+                        iterator = self.broker.listen()
+                        self.sem_prefetch.release()
                     continue
         finally:
             # We don't want to fetch new messages if we are shutting down.
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -383,7 +383,7 @@ class Receiver:
         if self.on_exit is not None:
             self.on_exit(self)
 
-    async def prefetcher(
+    async def prefetcher(  # noqa: C901
         self,
         queue: "asyncio.Queue[bytes | AckableMessage]",
         finish_event: asyncio.Event,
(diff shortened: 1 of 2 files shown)


s3rius:
Hello. Thanks for your contribution. However, I'd love to move `try` inside the while loop, because in case of any error prefetcher will stop working completely which will make worker to hang. It might happen, because of `asyncio.wait` for example.

PythonFZ:
> Hello. Thanks for your contribution. However, I'd love to move `try` inside the while loop, because in case of any error prefetcher will stop working completely which will make worker to hang. It might happen, because of `asyncio.wait` for example.

That is a good point, I added a broad `try ... except Exception` in https://github.com/taskiq-python/taskiq/pull/630/changes/592b167724cf8d014507cd733b950b52ee7fd75d and a test for transient listening failures in https://github.com/taskiq-python/taskiq/pull/630/changes/adeb379a4dbaf69ad6ee90867e3e442d8184505f

Does that work for you or are there any concerns with calling `iterator = self.broker.listen()` in the exception handling?

PR #634, opened 2026-06-16
feat: add worker-side task batching

KatantDev:
Allow grouping same-type tasks into batches that are executed together, reducing per-call overhead for tasks like DB writes, external API calls, ML inference, event publishing and indexing.

Usage:

```python
@broker.task(batch=True, batch_size=100, batch_timeout=3)
async def process_items(items: list[Item]) -> None:
    ...

await process_items.kiq(item)  # one item per kiq
```

Each .kiq sends a single item; the worker buffers messages per task name and invokes the function once with the collected list. A batch is flushed when batch_size is reached or batch_timeout seconds pass since the first buffered item (whichever comes first), and on graceful shutdown.

Details:
- AsyncBatchedTaskiqDecoratedTask types .kiq to accept a single element while the function body receives list[item].
- Batch config is validated at registration (TaskiqBatchConfigError).
- The whole batch shares one result, stored under every task_id, and every message is acked per the configured AcknowledgeType. pre_execute, post_execute, post_save and on_error middlewares run per message.
- InMemoryBroker supports batching too (flush on size, on wait_all, or immediately when await_inplace=True) for testing.

### DIFF
--- a/taskiq/abc/broker.py
+++ b/taskiq/abc/broker.py
@@ -10,6 +10,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     ClassVar,
+    Literal,
     ParamSpec,
     TypeAlias,
     TypeVar,
@@ -21,9 +22,9 @@ from uuid import uuid4
 from taskiq.abc.middleware import TaskiqMiddleware
 from taskiq.abc.serializer import TaskiqSerializer
 from taskiq.acks import AckableMessage
-from taskiq.decor import AsyncTaskiqDecoratedTask
+from taskiq.decor import AsyncBatchedTaskiqDecoratedTask, AsyncTaskiqDecoratedTask
 from taskiq.events import TaskiqEvents
-from taskiq.exceptions import TaskBrokerMismatchError
+from taskiq.exceptions import TaskBrokerMismatchError, TaskiqBatchConfigError
 from taskiq.formatters.proxy_formatter import ProxyFormatter
 from taskiq.message import BrokerMessage
 from taskiq.result_backends.dummy import DummyResultBackend
@@ -43,6 +44,7 @@ if TYPE_CHECKING:  # pragma: no cover
     from taskiq.abc.result_backend import AsyncResultBackend
 
 _T = TypeVar("_T")
+_Item = TypeVar("_Item")
 _FuncParams = ParamSpec("_FuncParams")
 _ReturnType = TypeVar("_ReturnType")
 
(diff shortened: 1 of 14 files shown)


taskiq/decor.py
```diff
@@ -232,3 +234,37 @@ def kicker(self) -> AsyncKicker[_FuncParams, _ReturnType]:
...
+        self,
+        item: _Item,
+    ) -> AsyncTaskiqTask[_ReturnType]:
+        """
+        Send a single item to be processed as part of a batch.
+
+        :param item: one element that becomes a member of the batched list.
+        :returns: taskiq task for this individual message.
+        """
+        kicker = cast(
```
s3rius, line 266: Why would you need this cast?

taskiq/abc/broker.py
```diff
@@ -255,6 +257,40 @@ def listen(self) -> AsyncGenerator[bytes | AckableMessage, None]:
...
+    @overload
+    def task(
+        self,
+        *,
+        batch: Literal[True],
+        batch_size: int | None = None,
+        batch_timeout: float | None = None,
+        **labels: Any,
+    ) -> Callable[
+        [Callable[[list[_Item]], Awaitable[_ReturnType]]],
```
s3rius, line 289: This definition ignores possible use of sync functions. Is it by design?

taskiq/abc/broker.py
```diff
@@ -255,6 +257,40 @@ def listen(self) -> AsyncGenerator[bytes | AckableMessage, None]:
...
+        if invalid:
+            raise TaskiqBatchConfigError
+
+    @overload
+    def task(
+        self,
+        *,
+        batch: Literal[True],
+        batch_size: int | None = None,
+        batch_timeout: float | None = None,
```
s3rius, line 286: You forgot `task_name` here.

taskiq/abc/broker.py
```diff
@@ -255,6 +257,40 @@ def listen(self) -> AsyncGenerator[bytes | AckableMessage, None]:
...
+        **labels: Any,
+    ) -> Callable[
+        [Callable[[list[_Item]], Awaitable[_ReturnType]]],
+        AsyncBatchedTaskiqDecoratedTask[_Item, ..., _ReturnType],
+    ]:  # pragma: no cover
```
s3rius, line 291: I don't think that we need this decorator class. Since it's used only for type annotations, I see that there is a more straightforward way for deconstructing ParamSpec. 

You can leverage the power of `Concatenate` and actually deconstruct ParamSpec to later reconstruct a desired variant.



```suggestion
    ) -> Callable[
        [Callable[Concatenate[list[_Item], _FuncParams], Awaitable[_ReturnType]]],
        AsyncTaskiqDecoratedTask[Concatenate[_Item, _FuncParams], _ReturnType],
    ]:  # pragma: no cover

```

This way, we ensure that first argument of the function is actually a list of items by deconstructing its type signature. It's kinda like pattern matching on types. 
And after that we reconstruct a desired form of ParamSpec by concatenating the rest of params with a first argument. 


However, this implementation allows for more than one argument. Which is might be not as preferable. Because it would complicate the batcher's logic a bit. But might be worth looking into.


If you want to support your original single item type, it can be easily done with the following signature:

 
```suggestion
    ) -> Callable[
        [Callable[[list[_Item]], Awaitable[_ReturnType]]],
        AsyncTaskiqDecoratedTask[[_Item], _ReturnType],
    ]:  # pragma: no cover
```

taskiq/abc/broker.py
```diff
@@ -332,8 +369,11 @@ def inner(
                 if "return" in sign:
                     return_type = sign["return"]
 
+                decorator_cls = self.decorator_class
+                if inner_labels.get("batch"):
```
s3rius, line 373: No need for custom decorator class as shown above.

s3rius (changes requested):
That's a good intention. And I like the implementation. However, there are few things about it.

First, I managed to find a way to make worker hang completely with this script:

<details>

<summary>Script to hang</summary>

```python
import asyncio

from taskiq import InMemoryBroker

broker = InMemoryBroker()


@broker.task(batch=True, batch_size=60, batch_timeout=3, task_name="tfunc")
async def tfunc(test: list[int]) -> bool:
    print(len(test))  # noqa: T201
    return True


async def main() -> None:
    await broker.startup()
    # Each `.kiq` sends a single item. They are buffered and run
    # together once the batch is flushed.
    tasks = await asyncio.gather(*[tfunc.kiq(i) for i in range(400)])
    # In tests, `wait_all` flushes any pending batches and waits
    # for them to finish before we read the results.
    await broker.wait_all()
    res = await asyncio.gather(*[task.wait_result(timeout=10) for task in tasks])
    len(res)
    await broker.shutdown()


if __name__ == "__main__":
    asyncio.run(main())
```

</details>

Also, this implementation does not support:

* More than one argument.
* Sync functions. 

I also left some comments.

PR #643, opened 2026-07-13
fix(inmemory): stabilize task ordering and lifecycle

### LINKED ISSUE #586: post_send fires after task execution with await_inplace=True
### Taskiq version

0.11.20

### Python version

Python 3.13

### OS

macos 26.2

### What happened?

When using InMemoryBroker(await_inplace=True), the post_send hook fires after the task completes instead of after the message is sent.

  Current behavior:
  pre_send → pre_execute → task → post_execute → post_send

  Expected behavior:
  pre_send → post_send → pre_execute → task → post_execute

  This breaks the semantic meaning of "post send". If I'm logging "task queued" in post_send, it ends up with a later timestamp than "task started" in pre_execute.

  Is this intentional? If so, might be worth documenting. If not, the broker could call post_send before invoking the task inline.

### Relevant log output

```shell

```

```python
from taskiq import InMemoryBroker, TaskiqMessage
  from taskiq.middlewares import SimpleRetryMiddleware
  from taskiq.abc.middleware import TaskiqMiddleware

  class TaskLoggingMiddleware(TaskiqMiddleware):
      async def _log_event(self, message: TaskiqMessage, status: str) -> None:
          # Inserts row to DB with created_at = now()
          # Problem: SENT rows end up with later timestamps than STARTED rows
          # because post_send fires after task execution with await_inplace=True
          pass

      async def post_send(self, message: TaskiqMessage) -> None:
          await self._log_event(message, "SENT")

      async def pre_execute(self, message: TaskiqMessage) -> TaskiqMessage:
          await self._log_event(message, "STARTED")
          return message

  broker = InMemoryBroker(await_inplace=True).with_middlewares(
      TaskLoggingMiddleware(),
      SimpleRetryMiddleware(default_retry_count=3),
  )
```

GefMar:
Ensure post_send completes before local execution while preserving per-invocation ownership and await_inplace semantics.

Drain accepted tasks during shutdown, retain background failures, and clean up middleware, result backend, and executor resources reliably.

Add concurrency, cancellation, lifecycle, and compatibility coverage.

Closes: #586
## Why This Approach Differs from #639

PR #639 correctly identifies the original problem and the required hook ordering. However, its implementation covers only the sequential happy path and does not establish reliable ownership of locally executed tasks.

### Per-Invocation Ownership

PR #639 stores inline tasks in a broker-wide `_inplace_tasks` set shared by all concurrent `.kiq()` calls. One invocation can therefore collect and await another invocation’s task, while the second invocation observes an empty set and returns before its own task completes.

This breaks the per-invocation contract of `await_inplace=True`.

The current implementation assigns each invocation its own execution task and synchronization gate. Every caller waits only for the work it submitted.

### Strict `post_send` Ordering

Creating the receiver with `asyncio.create_task()` before running `post_send` does not guarantee strict ordering. When an asynchronous `post_send` hook suspends, the event loop may start `pre_execute` and the task body concurrently.

The implementation in this PR keeps the receiver behind a per-invocation gate. Execution cannot begin until the complete `post_send` middleware chain for that invocation has finished.

### Error and Cancellation Ownership

If `post_send` fails or is cancelled after the message has been accepted, the execution task still needs an explicit owner.

This implementation transfers such work to broker-managed background ownership. `wait_all()` and `shutdown()` can then drain it and propagate execution failures instead of leaving detached tasks behind.

It also preserves the existing cancellation behavior for normal inline execution.

### Lifecycle Integration

The ownership model is integrated with the complete `InMemoryBroker` lifecycle:

- completed background failures remain observable until drained;
- cancelling `wait_all()` does not cancel broker-owned work;
- `shutdown()` drains accepted tasks before closing middleware, the result backend, and the executor;
- cleanup continues after failures while preserving the first raised exception;
- reentrant drain attempts fail explicitly instead of deadlocking.

### Test Coverage

The behavior is protected by deterministic tests covering:

- suspending asynchronous `post_send` hooks;
- concurrent `.kiq()` calls;
- `post_send` failures and cancellation;
- delayed background execution failures;
- overlapping sends and shutdown;
- direct `kick()` compatibility;
- unchanged behavior for regular external brokers.

The additional implementation is therefore not ceremony around hook ordering. It defines the complete lifecycle of an accepted message: gated execution, per-invocation or broker ownership, failure propagation, draining, and cleanup.

### DIFF
--- a/tests/abc/test_broker.py
+++ b/tests/abc/test_broker.py
@@ -4,9 +4,11 @@ from copy import copy
 import pytest
 
 from taskiq.abc.broker import AsyncBroker
+from taskiq.abc.middleware import TaskiqMiddleware
 from taskiq.decor import AsyncTaskiqDecoratedTask
 from taskiq.events import TaskiqEvents
-from taskiq.message import BrokerMessage
+from taskiq.exceptions import SendTaskError
+from taskiq.message import BrokerMessage, TaskiqMessage
 from taskiq.state import TaskiqState
 
 
@@ -28,6 +30,46 @@ class _TestBroker(AsyncBroker):
 
         :param callback: callback that is never called.
         """
+
+
+class _RecordingSendBroker(_TestBroker):
+    """Record transport dispatch and raise an optional send failure."""
+
+    def __init__(
+        self,
+        events: list[str],
+        kick_error: Exception | None = None,
+    ) -> None:
+        super().__init__()
+        self.events = events
+        self.kick_error = kick_error
+
+    async def kick(self, message: BrokerMessage) -> None:
+        self.events.append("kick")
+        if self.kick_error is not None:
+            raise self.kick_error
+
+
+class _RecordingSendMiddleware(TaskiqMiddleware):
+    """Record client send hooks and raise an optional post-send failure."""
+
+    def __init__(
+        self,
+        events: list[str],
+        post_send_error: Exception | None = None,
+    ) -> None:
+        super().__init__()
+        self.events = events
+        self.post_send_error = post_send_error
+
+    def pre_send(self, message: TaskiqMessage) -> TaskiqMessage:
+        self.events.append("pre_send")
+        return message
+
+    def post_send(self, message: TaskiqMessage) -> None:
+        self.events.append("post_send")
+        if self.post_send_error is not None:
+            raise self.post_send_error
 
 
 def test_decorator_success() -> None:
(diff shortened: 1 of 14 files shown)


tests/abc/test_broker.py
```diff
@@ -30,6 +35,79 @@ async def listen(self) -> AsyncGenerator[BrokerMessage, None]:  # type: ignore
         """
 
 
+class _RecordingSendBroker(_TestBroker):
+    """Record transport dispatch and raise an optional send failure."""
+
+    def __init__(
+        self,
+        events: list[str],
```
s3rius: Do you need a list here? I guess you can omit those arguements to set empty lists later in constructor.

tests/abc/test_broker.py
```diff
@@ -30,6 +35,79 @@ async def listen(self) -> AsyncGenerator[BrokerMessage, None]:  # type: ignore
...
+        if self.kick_error is not None:
+            raise self.kick_error
+
+
+class _RecordingSendMiddleware(TaskiqMiddleware):
+    """Record client send hooks and raise an optional post-send failure."""
+
+    def __init__(
+        self,
+        events: list[str],
```
s3rius: Same thing.

PR #645, opened 2026-07-15
fix(receiver): harden prefetch and listener lifecycle

### LINKED ISSUE #528: max-prefetch argument doesn't work as expected

GefMar:
Track delivery and prefetch capacity ownership across shutdown paths. Close pending listener reads, preserve late deliveries, and restore fail-fast transport error propagation.

Keep middleware accounting balanced, report discarded deliveries, and retain deferred lifecycle diagnostics when task group failures overlap.

Refs #528

im evaluated [PR#630](https://github.com/taskiq-python/taskiq/pull/630) and retained its valid requirement: the Receiver must not pull a message it cannot execute. 

However, that implementation also introduces a broad retry loop that treats transport, middleware, and programming errors alike, while pending-read, iterator, late-delivery, and semaphore ownership remain implicit.

This PR instead fixes the underlying ownership invariant. Each admitted delivery explicitly owns capacity from broker read through callback completion, pending reads are cancelled and awaited, iterators are closed, and failures remain visible to the worker lifecycle owner.

Retry and backoff remain a separate policy rather than being introduced implicitly as part of the prefetch fix.

### DIFF
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 from enum import Enum, auto
 from logging import getLogger
 from time import time
-from typing import Any, get_type_hints
+from typing import Any, Literal, get_type_hints
 
 import anyio
 from taskiq_dependencies import DependencyGraph
@@ -645,7 +645,13 @@ class Receiver:
         state.owns_delivery_slot = False
         return late_delivery
 
-    async def _notify_prefetch_hook(self, hook_name: str) -> None:
+    async def _notify_prefetch_hook(
+        self,
+        hook_name: Literal[
+            "on_prefetch_queue_add",
+            "on_prefetch_queue_remove",
+        ],
+    ) -> None:
         """Run all prefetch hooks and preserve the first failure."""
         first_error: BaseException | None = None
         for middleware in reversed(self.broker.middlewares):
@@ -723,7 +729,18 @@ class Receiver:
                         await asyncio.wait(tasks, timeout=self.wait_tasks_timeout)
                         logger.info("No more tasks to wait for. Shutting down.")
                     break
-                started_callback = await self._start_callback(queued_message)
+                execution_semaphore = self.sem
+                owns_execution_slot = execution_semaphore is not None
+                if execution_semaphore is not None:
+                    try:
+                        await execution_semaphore.acquire()
+                    except BaseException:
+                        queue.put_nowait(queued_message)
+                        raise
+                started_callback = await self._start_callback(
+                    queued_message,
+                    owns_execution_slot=owns_execution_slot,
+                )
                 tasks.add(started_callback.task)
 
                 # We want the task to remove itself from the set when it's done.
@@ -747,15 +764,13 @@ class Receiver:
     async def _start_callback(
         self,
         message: _PrefetchedMessage,
+        *,
+        owns_execution_slot: bool,
     ) -> _StartedCallback:
         """Transfer execution and delivery capacity to a callback task."""
         owns_delivery_slot = message.owns_delivery_slot
-        owns_execution_slot = False
         try:
             await self._notify_prefetch_hook("on_prefetch_queue_remove")
-            if self.sem is not None:
-                await self.sem.acquire()
-                owns_execution_slot = True
 
             if self.sem is None and owns_delivery_slot:
                 self.sem_prefetch.release()
@@ -767,6 +782,9 @@ class Receiver:
                 owns_delivery_slot=owns_delivery_slot,
             )
         except BaseException:
+            logger.warning(
+                "Discarding 1 prefetched delivery during Receiver cleanup.",
+            )
             if owns_delivery_slot:
                 self.sem_prefetch.release()
             if owns_execution_slot and self.sem is not None:
--- a/taskiq/receiver/receiver.py
+++ b/taskiq/receiver/receiver.py
@@ -4,8 +4,10 @@ import functools
 import inspect
 import random
 import sys
-from collections.abc import Callable
+from collections.abc import AsyncGenerator, Callable
 from concurrent.futures import Executor, ProcessPoolExecutor
+from dataclasses import dataclass
+from enum import Enum, auto
 from logging import getLogger
 from time import time
 from typing import Any, get_type_hints
@@ -26,7 +28,37 @@ from taskiq.utils import maybe_awaitable
 
 logger = getLogger(__name__)
 PY_VERSION = sys.version_info
-QUEUE_DONE = b"-1"
+
+
+class _QueueSignal(Enum):
+    """Control signals exchanged by the Receiver queue."""
+
+    DONE = auto()
+
+
+@dataclass(frozen=True, slots=True)
+class _PrefetchedMessage:
+    """A delivery and its admission-capacity ownership."""
+
+    data: bytes | AckableMessage
+    owns_delivery_slot: bool
+
+
+@dataclass(slots=True)
+class _PrefetchState:
+    """State owned by one broker listener."""
+
+    iterator: AsyncGenerator[bytes | AckableMessage, None]
+    current_message: asyncio.Task[bytes | AckableMessage] | None = None
+    owns_delivery_slot: bool = False
+
+
+@dataclass(frozen=True, slots=True)
+class _StartedCallback:
+    """A callback task and the delivery capacity transferred to it."""
+
+    task: asyncio.Task[Any]
+    owns_delivery_slot: bool
 
 
 def _execute_sync_task_in_executor(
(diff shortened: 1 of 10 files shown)


taskiq/receiver/receiver.py
```diff
@@ -386,19 +424,37 @@ async def listen(self, finish_event: asyncio.Event) -> None:  # pragma: no cover
...
+            finally:
+                with anyio.CancelScope(shield=True):
+                    await self._discard_queued_messages(queue)
+        except BaseException:
+            if self._listen_error is not None:
+                error = self._listen_error
+                logger.error(
+                    "A Receiver listener lifecycle error was recorded before "
+                    "the task group failed.",
+                    exc_info=(type(error), error, error.__traceback__),
```
s3rius: This one can be simplified.

```suggestion
                    "A Receiver listener lifecycle error was recorded before "
                    "the task group failed.\n%s" % error,
                    exc_info=True,
```

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,246 @@ async def prefetcher(
...
+            return None
+
+        late_delivery = _PrefetchedMessage(
+            data=late_message,
+            owns_delivery_slot=state.owns_delivery_slot,
+        )
+        state.owns_delivery_slot = False
+        return late_delivery
+
+    async def _notify_prefetch_hook(self, hook_name: str) -> None:
```
s3rius: Let's remove this function and whole `on_prefetch_queue_add` and `on_prefetch_queue_remove`. They are not part of middelware ABC and should not be used. 

I actually missed that thing when opentelemetry support was added. Let's remove this functionality, since it's internal implementation details. I don't think those functions should have existed.

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,246 @@ async def prefetcher(
...
-        fetched_tasks: int = 0
-        iterator = self.broker.listen()
-        current_message: asyncio.Task[bytes | AckableMessage] = asyncio.create_task(
-            iterator.__anext__(),  # type: ignore
-        )
+        try:
+            state = _PrefetchState(iterator=self.broker.listen())
+        except BaseException as exc:
+            self._record_listen_error(exc)
+            queue.put_nowait(_QueueSignal.DONE)
```
s3rius: Since we're running in a anyio group, I suppose you can simply do something like this:


```suggestion
            queue.put_nowait(_QueueSignal.DONE)
            raise
```

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,246 @@ async def prefetcher(
...
+                    break
+        finally:
+            logger.info("Stopping prefetching messages...")
+            with anyio.CancelScope(shield=True):
+                try:
+                    await self._enqueue_late_prefetched_message(queue, state)
+                finally:
+                    queue.put_nowait(_QueueSignal.DONE)
+                    finish_waiter.cancel()
+                    await asyncio.gather(finish_waiter, return_exceptions=True)
```
s3rius:
```suggestion
                    with contextlib.supress(asyncio.CanceledError):
                        await finish_waiter
```

Idk, IMO it's just more obvious what is going on with such code.

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,246 @@ async def prefetcher(
...
+        try:
+            done, _ = await asyncio.wait(
+                {acquire_task, finish_waiter},
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+        except BaseException:
+            await self._settle_delivery_acquire(acquire_task)
+            raise
+
+        if finish_waiter in done or finish_event.is_set():
```
s3rius:
```suggestion
        if finish_waiter.done() or finish_event.is_set():
```

I guess that'd work and will look a bit cleaner. Also, you won't need done.

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,252 @@ async def prefetcher(
...
+        if first_error is not None:
+            raise first_error
+
+    async def _discard_queued_messages(
+        self,
+        queue: "asyncio.Queue[_PrefetchedMessage | _QueueSignal]",
+    ) -> None:
+        """Release capacity and instrumentation for abandoned queue entries."""
+        discarded_messages = 0
+        while True:
```
s3rius:
```suggestion
        while queue:
```

You can use this condition to simplify inner body of the loop. Because you won't need to check on QueueEmpty exception.

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,252 @@ async def prefetcher(
...
-                        )
+                queue.put_nowait(
+                    _PrefetchedMessage(
+                        data=message,
+                        owns_delivery_slot=True,
+                    ),
+                )
+                state.owns_delivery_slot = False
+                try:
+                    await self._notify_prefetch_hook("on_prefetch_queue_add")
```
s3rius: This can also be removed.

taskiq/receiver/receiver.py
```diff
@@ -407,54 +463,252 @@ async def prefetcher(
...
+    ) -> None:
+        """Run all prefetch hooks and preserve the first failure."""
+        first_error: BaseException | None = None
+        for middleware in reversed(self.broker.middlewares):
+            hook = getattr(middleware, hook_name, None)
+            if hook is not None:
+                try:
+                    await maybe_awaitable(hook())
+                except BaseException as exc:
+                    if first_error is None:
```
s3rius: This approach ignores all the errors except the first one. I mean, it's fine, because you raise it anyway. But, I'd rather create an [exception group](https://docs.python.org/3.11/library/exceptions.html#exception-groups). 

I think it's safe to bump MPE to 3.11 and start using ExceptionGroup for such case. It'd be a good fit and reasonable upgrade. 

Since python 3.10 will be completely discarded in 3 months.

s3rius (changes requested):

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: FIX

taskiq/receiver/receiver.py
GefMar: Using ExceptionGroup no longer provides any practical benefit. However, adopting it would change the exception contract and require dropping support for Python 3.10, which is still explicitly supported by the project and is part of the CI matrix.

Therefore, raising the project’s minimum supported Python version is outside the scope of this change.

I’m not in favor of introducing changes that conflict with the project’s current dependency constraints. We can revisit this as part of a technical debt initiative once support for Python 3.10 has been officially dropped.

s3rius: Well. It does make sense. Okay, let's leave it that way for now.

s3rius (approved):

PR #648, opened 2026-07-17
fix(cli): correct --hardkill-count off-by-one in worker signal handler

### LINKED ISSUE #647: --hardkill-count appears off by one
### Taskiq version

0.12.4

### Python version

Python 3.12

### OS

Linux

## Description

The documentation describes `--hardkill-count` as the number of termination signals allowed before performing a hardkill.

However, the counter is currently implemented as:

```python
hardkill_counter = 0

def interrupt_handler(signum, _frame):
    nonlocal hardkill_counter

    shutdown_event.set()

    if hardkill_counter > args.hardkill_count:
        logger.warning("Hard kill. Exiting.")
        raise KeyboardInterrupt

    hardkill_counter += 1
```

Because the comparison happens before incrementing and uses `>`, `--hardkill-count 3` behaves as follows:

| Signal | Counter before signal | Result |
|---:|---:|---|
| 1 | 0 | Graceful shutdown |
| 2 | 1 | Graceful shutdown |
| 3 | 2 | Graceful shutdown |
| 4 | 3 | Graceful shutdown |
| 5 | 4 | Hardkill |

If the value means “three termination signals before hardkill,” I would expect the fourth signal to trigger the hardkill. The current implementation triggers it on the fifth signal.

Would this condition be expected to use `>=` instead?

```python
if hardkill_counter >= args.hardkill_count:
    logger.warning("Hard kill. Exiting.")
    raise KeyboardInterrupt
```

gitcommit90:
## Summary
Fixes #647: `--hardkill-count N` was off by one in the worker signal handler.

Docs/CLI help describe N as the number of termination signals allowed before a hard kill (hard kill on signal N+1). The handler compared `hardkill_counter > args.hardkill_count` **before** incrementing, so with the default `N=3` the hard kill fired on the **5th** signal instead of the 4th.

## Changes
- `taskiq/cli/worker/run.py`: change `>` to `>=` so hard kill runs on signal N+1.
- Add `tests/cli/worker/hanging_broker.py` and `tests/cli/worker/test_hardkill_count.py` — integration test that spawns a real worker with a broker whose graceful shutdown hangs, sends SIGINTs to the worker child, and asserts survival through N signals and hard kill on N+1.

## Test plan
- [x] New hardkill-count integration test fails on pre-fix code and passes with the fix
- [x] Full suite: `pytest tests/ -q -n auto` → 298 passed
- [x] `ruff check`, `ruff format --check`, `black --check`, `mypy` clean on touched files

Closes #647

### DIFF
--- a/taskiq/cli/worker/run.py
+++ b/taskiq/cli/worker/run.py
@@ -110,7 +110,7 @@ def start_listen(args: WorkerArgs) -> None:
         shutdown_event.set()
         # Hard kill is a signal that we should stop
         # everything immediately.
-        if hardkill_counter > args.hardkill_count:
+        if hardkill_counter >= args.hardkill_count:
             logger.warning("Hard kill. Exiting.")
             raise KeyboardInterrupt
         hardkill_counter += 1
--- a/tests/cli/worker/test_hardkill_count.py
+++ b/tests/cli/worker/test_hardkill_count.py
@@ -1,0 +1,119 @@
+"""Integration test for --hardkill-count signal semantics.
+
+Spawns a real worker subprocess whose broker never finishes graceful
+shutdown, sends SIGINTs directly to a worker child process (the process
+that installs the hard-kill counting signal handler), and checks how
+many signals it takes to hard-kill it.
+
+Documented behavior (``--hardkill-count N``): N termination signals are
+allowed before performing a hardkill, i.e. signals 1..N only set the
+shutdown event and signal N+1 raises KeyboardInterrupt in the worker.
+"""
+
+import os
+import signal
+import subprocess
+import sys
+import time
+from pathlib import Path
+
+import psutil
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+SIGNAL_SETTLE_SECONDS = 0.6
+WAIT_TIMEOUT = 15.0
+
+
+def _pid_alive(pid: int) -> bool:
+    """Check whether a process exists."""
+    try:
+        os.kill(pid, 0)
+    except OSError:
+        return False
+    return True
+
+
+def _worker_child_pids(main_pid: int) -> list[int]:
+    """Return pids of worker child processes of the main process."""
+    try:
+        parent = psutil.Process(main_pid)
+    except psutil.NoSuchProcess:
+        return []
+    return [child.pid for child in parent.children(recursive=False)]
+
+
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="SIGINT hard-kill semantics differ on Windows",
+)
+def test_hardkill_count_signals_before_hardkill() -> None:
+    """N signals are graceful; signal N+1 hard-kills the worker process."""
+    hardkill_count = 3
+    env = os.environ.copy()
+    env["PYTHONPATH"] = str(REPO_ROOT) + os.pathsep + env.get("PYTHONPATH", "")
+    proc = subprocess.Popen(  # noqa: S603
+        [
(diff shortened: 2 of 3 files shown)


GefMar (commented):
A few notes on the signal semantics and test reliability.

taskiq/cli/worker/run.py
```diff
@@ -110,7 +110,7 @@ def interrupt_handler(signum: int, _frame: Any) -> None:
         shutdown_event.set()
         # Hard kill is a signal that we should stop
         # everything immediately.
-        if hardkill_counter > args.hardkill_count:
+        if hardkill_counter >= args.hardkill_count:
```
GefMar, line 113: Before changing this condition, could we clarify the contract? With this change, `--hardkill-count 3` kills the worker on the fourth signal, while the docs say that sending three signals performs a hard kill. We should pick one interpretation and align the code, CLI help, docs, and test.

tests/cli/worker/test_hardkill_count.py
```diff
@@ -0,0 +1,119 @@
...
+        deadline = time.monotonic() + WAIT_TIMEOUT
+        worker_pid: int | None = None
+        while time.monotonic() < deadline:
+            if proc.poll() is not None:
+                pytest.fail(
+                    f"Worker exited during startup: returncode={proc.returncode}",
+                )
+            children = _worker_child_pids(proc.pid)
+            if children:
+                worker_pid = children[0]
```
GefMar, line 88: On macOS, the first child is usually `multiprocessing.resource_tracker`, not the worker. It ignores `SIGINT`, so this test will fail there. We need a reliable way to get the actual worker PID.

tests/cli/worker/test_hardkill_count.py
```diff
@@ -0,0 +1,119 @@
...
+                break
+            time.sleep(0.25)
+        if worker_pid is None:
+            pytest.fail("No worker child process spawned in time")
+        # Extra settle time so the child's signal handlers are installed.
+        time.sleep(2.0)
```
GefMar, line 94: A fixed two-second sleep does not guarantee that the worker has installed its signal handlers, especially on a busy CI runner. Could we wait for an explicit readiness signal from the test broker instead?

tests/cli/worker/test_hardkill_count.py
```diff
@@ -0,0 +1,119 @@
...
+        while time.monotonic() < deadline and _pid_alive(worker_pid):
+            time.sleep(0.1)
+        assert not _pid_alive(worker_pid), (
+            f"Worker still alive after {hardkill_count + 1} signals; "
+            f"--hardkill-count {hardkill_count} should hard-kill on "
+            f"signal {hardkill_count + 1}"
+        )
+    finally:
+        if proc.poll() is None:
+            proc.kill()
```
GefMar, line 118: `proc.kill()` only stops the supervisor, so the hanging worker may be left running after a test failure. It would be safer to terminate the whole process group or explicitly clean up all child processes.

tests/cli/worker/test_hardkill_count.py
s3rius: What if in _worker_child_pids we will filter out all processes that are not `python`?

tests/cli/worker/test_hardkill_count.py
s3rius: We already set asyncio.event in the broker. I suppose, we can reuse it to fit our needs. Like, we only start killing, when the event is set. Reseting in on every iteration.

PR #651, opened 2026-07-24
feat: add per-task types_of_exceptions

spikeninja:


s3rius (dismissed):

s3rius:
Can you please test out what happens when you send a task with a tuple of exceptions in labels using some distributed brokers? I wonder what type will it end up with?

taskiq/middlewares/smart_retry_middleware.py
```diff
@@ -68,6 +68,28 @@ def __init__(
...
+
+        Per-task ``types_of_exceptions`` set via the task decorator take
+        precedence over the broker-wide value. Types are read from the
+        registered task object, since label values are stringified when a
+        message is serialized and cannot carry real exception types.
+
+        :param message: Original task message.
+        :return: Effective exception types or None.
+        """
+        task = self.broker.find_task(message.task_name)
```
GefMar, line 86: A regular label belongs to a specific message. It is produced by the producer, passed through the broker, and can be overridden via .kicker().with_labels(...).

In contrast, types_of_exceptions is effectively part of the task definition on the worker side. The value passed with the message is not used for any decision-making.


It seems to me that this results in the task labels being silently lost before the retry.

The same issue exists in simple_retry_middleware.py:44

GefMar (approved):

s3rius (approved):

PR #653, opened 2026-07-28
fix: allow registering bound methods as tasks

### LINKED ISSUE #436: No longer able to register methods in TaskIQ 0.11.10
After updating to `0.11.10`, registering a method as a task produces an `AttributeError`:

```
bbot_server/applets/_base.py:234: in register_watchdog_tasks
    task = broker.register_task(method, **kwargs)
../../../.cache/pypoetry/virtualenvs/bbot-server-DmHQwTaK-py3.13/lib/python3.13/site-packages/taskiq/abc/broker.py:377: in register_task
    return self.task(task_name=task_name, **labels)(func)
../../../.cache/pypoetry/virtualenvs/bbot-server-DmHQwTaK-py3.13/lib/python3.13/site-packages/taskiq/abc/broker.py:331: in inner
    self.decorator_class(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = AsyncTaskiqDecoratedTask(bbot_server.applets.events:archive_events_task), broker = <taskiq_nats.broker.NatsBroker object at 0x70a16466cec0>
task_name = 'bbot_server.applets.events:archive_events_task'
original_func = <bound method EventsApplet.archive_events_task of <bbot_server.applets.events.EventsApplet object at 0x70a16609c050>>, labels = {'schedule': [{'cron': '0 0 * * *'}]}

    def __init__(
        self,
        broker: "AsyncBroker",
        task_name: str,
        original_func: Callable[_FuncParams, _ReturnType],
        labels: Dict[str, Any],
    ) -> None:
        self.broker = broker
        self.task_name = task_name
        self.original_func = original_func
        self.labels = labels
    
        # This is a hack to make ProcessPoolExecutor work
        # with decorated functions.
        #
        # The problem is that when we decorate a function
        # it becomes a new class. This class has the same
        # name as the original function.
        #
        # When receiver sends original function to another
        # process, it will have the same name as the decorated
        # class. This will cause an error, because ProcessPoolExecutor
        # uses `__name__` and `__qualname__` attributes to
        # import functions from other processes and then it verifies
        # that the function is the same as the original one.
        #
        # This hack renames the original function and injects
        # it back to the module where it was defined.
        # This way ProcessPoolExecutor will be able to import
        # the function by it's name and verify its correctness.
        new_name = f"{original_func.__name__}__taskiq_original"
>       self.original_func.__name__ = new_name
E       AttributeError: 'method' object has no attribute '__name__' and no __dict__ for setting new attributes. Did you mean: '__ne__'?

../../../.cache/pypoetry/virtualenvs/bbot-server-DmHQwTaK-py3.13/lib/python3.13/site-packages/taskiq/decor.py:79: AttributeError
```

sescer:
## Summary
- Skip ProcessPoolExecutor `__name__`/`__qualname__` rewriting unless `original_func` is a plain Python function (`inspect.isfunction`)
- Restore `broker.register_task(instance.method, ...)` for bound methods (#436)
- Add regression tests for registration, execution with `self`, and ProcessPoolExecutor pickle/import of module-level sync tasks

Fixes #436

### DIFF
--- a/tests/abc/test_broker.py
+++ b/tests/abc/test_broker.py
@@ -1,8 +1,11 @@
 from collections.abc import AsyncGenerator
+from concurrent.futures import ProcessPoolExecutor
 from copy import copy
+from types import MethodType
 
 import pytest
 
+from taskiq import InMemoryBroker
 from taskiq.abc.broker import AsyncBroker
 from taskiq.decor import AsyncTaskiqDecoratedTask
 from taskiq.events import TaskiqEvents
@@ -28,6 +31,15 @@ class _TestBroker(AsyncBroker):
 
         :param callback: callback that is never called.
         """
+
+
+_process_pool_broker = _TestBroker()
+
+
+@_process_pool_broker.task(task_name="process_pool_sync_add")
+def process_pool_sync_add(value: int) -> int:
+    """Module-level sync task used by ProcessPoolExecutor regression test."""
+    return value + 1
 
 
 def test_decorator_success() -> None:
@@ -80,6 +92,53 @@ def test_kicker_labels_modification() -> None:
     assert "another_label" in test_kicker.labels
 
     assert test_task.labels == old_labels
+
+
+def test_register_task_accepts_bound_method() -> None:
+    """Bound methods must register without mutating method.__name__."""
+    broker = _TestBroker()
+
+    class Counter:
+        def increment(self, value: int) -> int:
+            return value + 1
+
+    instance = Counter()
+    task = broker.register_task(instance.increment)
+
+    assert isinstance(task, AsyncTaskiqDecoratedTask)
+    assert isinstance(task.original_func, MethodType)
+    assert task.original_func.__self__ is instance
+    assert task.task_name == f"{instance.increment.__module__}:increment"
+    assert broker.find_task(task.task_name) is task
+
+
+@pytest.mark.anyio
+async def test_bound_method_task_executes_with_self() -> None:
+    """Registered bound method must keep instance state across execution."""
+    broker = InMemoryBroker(await_inplace=True)
+
+    class Counter:
+        def __init__(self) -> None:
+            self.total = 0
+
+        def add(self, value: int) -> int:
+            self.total += value
+            return self.total
+
+    instance = Counter()
+    task = broker.register_task(instance.add, task_name="bound.add")
+
+    kicked = await task.kiq(5)
+    result = await kicked.wait_result()
+    assert result.return_value == 5
+    assert instance.total == 5
+
+
+def test_process_pool_runs_module_level_decorated_sync_function() -> None:
+    """ProcessPoolExecutor must be able to run renamed original sync functions."""
+    with ProcessPoolExecutor(max_workers=1) as executor:
+        future = executor.submit(process_pool_sync_add.original_func, 41)
+        assert future.result(timeout=5) == 42
 
 
 @pytest.mark.anyio
--- a/taskiq/decor.py
+++ b/taskiq/decor.py
@@ -1,3 +1,4 @@
+import inspect
 import sys
 from collections.abc import Callable, Coroutine
 from copy import copy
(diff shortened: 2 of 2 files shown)


tests/abc/test_broker.py
```diff
@@ -82,6 +94,53 @@ async def test_task() -> None: ...
...
+def test_register_task_accepts_bound_method() -> None:
+    """Bound methods must register without mutating method.__name__."""
+    broker = _TestBroker()
+
+    class Counter:
+        def increment(self, value: int) -> int:
+            return value + 1
+
+    instance = Counter()
+    task = broker.register_task(instance.increment)
```
s3rius, line 106: There is a problem with this approach and it should not be allowed to do so. 

You register a task as a function which is bound to a particular instance of the class. The problem is that there is a hidden state in the class.

For example, consider this:


```python
class Sateful:
    def __init__(self, inc: int) -> None:
		self.inc = inc
	
	def my_task(self, value: int) -> int:
		return self.value + self.inc

instance = Counter()
task = broker.register_task(instance.my_task)
```

There are few questions to be answered:
* How do we make sure that the state of the class matches the state of the task?
* What to do in case if it doesn't?
* It it's an instance, what should be the name of such function?
* If there are multiple instances of the same class with different state, how do we handle it?

s3rius (commented):
Hello! Thanks for your contribution, but I have few questions about your approach.

tests/abc/test_broker.py
sescer: Hello! Thanks for the review -  these are fair design questions!

_Quick disclaimer: this is my first open-source contribution in a long time, so I may be missing some project context or conventions. Happy to adjust based on your guidance._

I agree that registering a live bound method is problematic: it captures hidden instance state that is not part of the task message. Taskiq’s model is closer to "lookup callable by name on the worker", so it cannot guarantee that client-side `self` matches worker-side `self`.

**How do we make sure class state matches the task?**

> With bound methods -  we don’t, at least not automatically. The app would have to register the same object on the worker during startup/import. That is fragile for distributed workers.

**What to do if it doesn’t?**

> Same as today for a missing/wrong registration: task not found, or the wrong callable runs. I don’t think core should serialize `self` into the message or try to validate instance identity across processes.

**What should the name be for an instance method?**
> Current default `module:__name__` is fine for functions. For instance methods it becomes ambiguous if several instances share the same method name. Explicit `task_name=...` would be required, but that still doesn’t solve state sync.

**Multiple instances with different state?**
> They would collide on the same default name, and last registration wins. Auto-namespacing by id(instance) would be unstable across processes, so I wouldn’t go that way.


So I don’t think “officially support register_task(instance.method)” is the right long-term answer for the project

sescer: Maybe we can do instead, if OOP-style tasks are desirable, is class-based tasks or maybe some sort of factory-based tasks:

- register a class, not a live instance
- worker instantiates it per execution, like `MyTask(broker, message).run(*args, **kwargs)`
- instance state is created on the worker (config, DI, labels), not captured from the client
- task name is stable: `module:ClassName` or explicit task_name;
- multiple logical variants are different classes / different names / different kwargs, not different live objects

That removes the hidden-state problem and fits the existing registry model much better than bound methods.

sescer: I would like to get some answers from you or other people with project context.

- For bound methods short-term, do you prefer: reject them with a clear error or keep no-crash registration only as unsupported behavior?
- Are you open to class-based tasks as a follow-up feature?
- If yes to class-based tasks, what should the minimal API look like to you: like `@broker.task` on a class with `run(...)`, or explicit M`yTask.register(broker)`, or idk....
- Should task instance state come mainly from constructor args / factory or  DI or some meta or ...

Happy to reshape this PR once you say which direction you want)

sescer: @s3rius )

PR #654, opened 2026-07-31
feat: forward broker-specific acknowledgement options

vvanglro:
## Summary

- Allow `Context.ack(**kwargs)` to forward broker-specific acknowledgement options to the broker callback.
- Update `AckController` and `AckableMessage` callback typing for optional keyword options.
- Document the callback contract for broker implementers.
- Add unit and receiver coverage for forwarding `delete_after_ack=True` through a manual acknowledgement.

## Compatibility

Automatic acknowledgements still invoke callbacks without options. Existing brokers with no-argument acknowledgement callbacks therefore continue to work unchanged unless an application explicitly passes broker-specific options through `Context.ack(...)`.

Taskiq intentionally does not define or interpret any option names; their semantics remain broker-specific.

## Validation

```bash
ruff check taskiq/acks.py taskiq/context.py tests/test_acks.py tests/receiver/test_receiver.py
python -m mypy taskiq/acks.py taskiq/context.py tests/test_acks.py tests/receiver/test_receiver.py
python -m pytest tests/test_acks.py tests/receiver/test_receiver.py -q
```

Result: `38 passed`.

### DIFF
--- a/docs/extending-taskiq/broker.md
+++ b/docs/extending-taskiq/broker.md
@@ -36,12 +36,18 @@ async def listen(self) -> AsyncGenerator[AckableMessage, None]:
    for message in self.my_channel:
       yield AckableMessage(
          data=message.bytes,
-         # Ack is a function that takes no parameters.
-         # So you either set here method of a message,
+         # Ack can accept broker-specific keyword options.
+         # So you either set here a method of a message,
          # or you can make a closure.
          ack=message.ack,
       )
 ```
+
+For manual acknowledgement, options passed to `Context.ack(**kwargs)` are
+forwarded unchanged to this callback. Taskiq does not define or interpret these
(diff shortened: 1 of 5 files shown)


vvanglro:
The original intention of doing this PR was that when I was making the ack for [taskiq-redis-streams](https://github.com/vvanglro/taskiq-redis-streams/commit/a2bdb142a492447a8ae808cef471b042df362da4#diff-13cc0395fa93c6e4ccc27e2ab534de6d9cd3416a4d21d70d7478fd805f335e0aL327) , I thought it would be possible to control the granularity of whether to delete this ack message.

But just now, while I was taking a shower, I brainstormed and it seems that users don't care about when the messages in the broker are deleted after the task is completed. Therefore, I am closing this PR.

PR #661, opened 2026-08-24
fix: ValueError "Token was created in a different context"

aticie:
Fixes an issue that happens with `OpenTelemetryMiddleware` when a message is requeued with `Context.requeue()`. 

`Context.requeue()` uses `self.broker.kick` to kick a message to the broker which skips the Middleware `pre-send` and `post-send` invocations that kicker does here: https://github.com/taskiq-python/taskiq/blob/ced1909574128687c6dd0d09976c6eb0131ab085/taskiq/kicker.py#L160-L162
 https://github.com/taskiq-python/taskiq/blob/ced1909574128687c6dd0d09976c6eb0131ab085/taskiq/kicker.py#L168-L170

Messages that are requeued with the same context variables will raise an error on `post_save`'s `.detach()` because they are now running in a different async context. 

The message's lifecycle on OpenTelemetryMiddleware will be:

- `pre_send` -> `message.labels` are injected with context here
- `post_send`
- `pre_execute` -> `message.labels` are extracted here and re-used
- `post_execute`
- `requeue()` happens -> skipping `pre_send` and `post_send`, therefore not renewing the context.
- `pre_execute` -> context inferred from `message.labels` again
- `post_execute`
- `post_save` -> Detaching the stale context here. Raises `ValueError: Token was created in a different context`

This change gets rid of `post_save` on OpenTelemetryMiddleware to detach the context on `post_execute` instead. Since, we are not actually doing anything database save related on the current `post_save` method, it seems fair to move everything under `post_execute` as this is where the task execution actually ends.

### DIFF
--- a/taskiq/middlewares/opentelemetry_middleware.py
+++ b/taskiq/middlewares/opentelemetry_middleware.py
@@ -312,40 +312,6 @@ class OpenTelemetryMiddleware(TaskiqMiddleware):
         )
         return message
 
-    def post_save(  # pylint: disable=R6301
-        self,
-        message: TaskiqMessage,
-        result: TaskiqResult[T],
-    ) -> None:
-        """
-        This function closes span from `pre_execute`.
-
-        :param message: received message.
-        :param result: result of the execution.
-        """
-        logger.debug("post_execute task_id=%s", message.task_id)
-
-        # retrieve and finish the Span
-        ctx = retrieve_context(message)
-
-        if ctx is None:
-            logger.warning("no existing span found for task_id=%s", message.task_id)
-            return
-
-        span, activation, token = ctx
-
-        if span.is_recording():
-            span.set_attribute(_TASK_TAG_KEY, _TASK_EXECUTE)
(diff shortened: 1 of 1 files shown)


danfimov:
@soapun Can you help me with review here? I remeber that you built opentelemetry integration. Maybe you have an opinion on this change. Thank you)

s3rius (approved):
This fix for the situation is actually correct. 

The only 1 concern I have is that we will skip recording save events. In context of the system it might be okay.

s3rius:
@aticie, but can you please run `prek` locally and fix issues? I rebased your PR onto new changes.

soapun:
> This fix for the situation is actually correct.
> 
> The only 1 concern I have is that we will skip recording save events. In context of the system it might be okay.

Agree on the concern. 

I think requeued task's span must be in the same trace as the original task span or even its child. Not sure if this is current behaviour. I guess it would be nice to be able to trigger `OpenTelemetryMiddleware` on `requeue`.

PR #667, opened 2026-08-30
fix: align positional argument parsing with parameter positions

### LINKED ISSUE #666: Worker misparses positional arguments when a preceding parameter lacks a type annotation
### Taskiq version

Taskiq version `0.12.6 (master)`

### Python version

Python 3.12

### OS

Linux

### What happened?

`parse_params` in `taskiq/receiver/params_parser.py` maps positional arguments to annotations with a counter (`argnum`) that only increments for annotated parameters. Unannotated parameters are skipped with `continue` before `argnum += 1`, but `get_type_hints()` omits them too, so they still occupy a slot in `message.args`. When any positional parameter has no annotation, every argument after it is parsed against the wrong annotation.

Example:

```python
@broker.task
async def my_task(request_id, count: int) -> None: ...

await my_task.kiq("12345", 3)
```

What the worker receives (verified on master, taskiq 0.12.6):

1. `request_id` arrives as the `int` `12345`. The unannotated parameter's value is coerced by the next parameter's annotation.
2. `count` is never validated. `kiq("hello", "not-an-int")` delivers `"not-an-int"` to `count`, and the only log line is a `Can't parse argument 0` warning about the other parameter.
3. If the first value fails to parse, the annotated parameter after it never gets coerced either. It keeps its wire representation, so a `datetime` parameter receives an ISO string.
4. `*values: int` varargs coerce only the first element. `kiq("1", "2", "3")` arrives as `[1, "2", "3"]`.

Standalone repro of the worker path, no broker needed:

```python
import inspect
from typing import get_type_hints

from taskiq.formatters.json_formatter import JSONFormatter
from taskiq.message import TaskiqMessage
from taskiq.receiver.params_parser import parse_params

fmt = JSONFormatter()

def my_task(request_id, count: int) -> None: ...

msg = fmt.loads(fmt.dumps(TaskiqMessage(
    task_id="1",
    task_name="mod:my_task",
    labels={},
    labels_types={},
    args=["12345", 3],
    kwargs={},
)).message)
parse_params(inspect.signature(my_task), get_type_hints(my_task), msg)
print(msg.args)  # [12345, 3] - request_id silently became an int
```

Each positional argument should be parsed against the annotation of the parameter in the same position, and unannotated parameters should be left untouched. Mixing untyped and typed positional parameters (a context or config object as the first argument is a common case) hits this without any error.

### Relevant log output

```shell
With `kiq("hello", "not-an-int")` the worker logs a warning for argument 0, the unannotated parameter, and nothing for `count`:


Can't parse argument 0 for task mod:my_task. Reason: 1 validation error for int
  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='hello', input_type=str]
```

```python
Not needed. The repro exercises the receiver's parsing path directly.
```

juanmicl:
Closes: #666

`parse_params` mapped positional arguments to annotations using a counter
that only advanced for annotated parameters, so any unannotated
positional parameter shifted every following argument onto the wrong
annotation (values silently coerced by the wrong type, annotated
parameters never validated).

This PR decouples the positional slot index from annotation presence:

- every `POSITIONAL_ONLY` / `POSITIONAL_OR_KEYWORD` parameter advances
  the slot index, annotated or not;
- parsing only happens when the parameter actually has an annotation;
- `*args` annotations now coerce **all** remaining positional arguments
  (previously only the first element was coerced);
- keyword-only parameters go straight to the kwargs branch.

Behavior for fully-annotated signatures is unchanged, all existing
`test_params_parser.py` cases pass as-is.

Validation: full suite passes (`pytest -q`, 313 tests), `black`, `ruff`
and `mypy` clean (mypy reports the same 2 pre-existing errors in
`taskiq/serializers/` as on master).

### DIFF
--- a/taskiq/receiver/params_parser.py
+++ b/taskiq/receiver/params_parser.py
@@ -8,6 +8,62 @@ from taskiq.message import TaskiqMessage
 logger = getLogger(__name__)
 
 
+def _parse_arg(
+    param_name: str,
+    annot: Any,
+    argnum: int,
+    message: TaskiqMessage,
+) -> None:
+    """
+    Parse a positional argument by its annotation, in place.
+
+    :param param_name: name of the parameter.
+    :param annot: type annotation of the parameter.
+    :param argnum: index of the argument in message.args.
+    :param message: incoming message.
+    """
+    value = message.args[argnum]
+    if value is None:
+        return
+    logger.debug("Trying to parse %s as %s", param_name, annot)
+    try:
+        # trying to parse found value as in type annotation.
+        message.args[argnum] = parse_obj_as(annot, value)
+    except (ValueError, RuntimeError) as exc:
+        logger.warning(
+            "Can't parse argument %d for task %s. Reason: %s",
+            argnum,
+            message.task_name,
+            exc,
+            exc_info=True,
+        )
+
+
+def _parse_kwarg(param_name: str, annot: Any, message: TaskiqMessage) -> None:
+    """
+    Parse a keyword argument by its annotation, in place.
+
+    :param param_name: name of the parameter.
+    :param annot: type annotation of the parameter.
+    :param message: incoming message.
+    """
+    value = message.kwargs.get(param_name)
+    if value is None:
+        return
+    logger.debug("Trying to parse %s as %s", param_name, annot)
+    try:
+        # trying to parse found value as in type annotation.
+        message.kwargs[param_name] = parse_obj_as(annot, value)
+    except (ValueError, RuntimeError) as exc:
+        logger.warning(
+            "Can't parse argument %s for task %s. Reason: %s",
+            param_name,
+            message.task_name,
+            exc,
+            exc_info=True,
+        )
+
+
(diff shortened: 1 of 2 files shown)


s3rius:
Okay, that seems like it might some the problem. But would you mind adding some tests to validate the fix?

juanmicl:
Added 9 tests to `tests/receiver/test_params_parser.py` covering the cases from the issue:

- unannotated parameters are left untouched (`request_id` keeps its string value)
- annotated parameters after an unannotated one are parsed in their own position (int and datetime)
- the `Can't parse argument N` warning now points at the right index
- `*args` coercion applies to all elements, not just the first
- keyword-only params and positional params passed as kwargs

Six of them fail on master and pass with the fix. Full suite passes locally (322 tests), ruff/black/mypy clean.

s3rius (approved):

s3rius:
Thanks for the contribution. Will appear in the next release.

PR #627, opened 2026-05-19
feat(scheduler): add optional send_timeout to prevent silent stall on hung kick

Rohit-Ekbote:
## Summary

Adds an optional `send_timeout` (CLI `--send-timeout`) on `SchedulerLoop.run` that wraps each spawned send task in `asyncio.wait_for`, so a hung `broker.kick` cannot indefinitely block subsequent ticks for the same `schedule_id` via the `running_schedules` skip check.

Default is `None` — no timeout, fully backwards-compatible.

## Problem

`SchedulerLoop.run` skips ticks for any `schedule_id` whose previous send task is still in `running_schedules`. The entry only leaves `running_schedules` via the task's `add_done_callback`, which fires only when the spawned coroutine completes. If `broker.kick` ever hangs (sentinel failover, half-open socket, network blip without TCP RST), the spawned send task never completes, the done_callback never fires, and **every subsequent tick for that schedule is silently skipped** — no error, no log, no recovery until process restart.

We hit this in production on `taskiq==0.11.20` (`ListQueueSentinelBroker` + redis-sentinel running through a flap). After upgrading to validate, the same logic is unchanged at HEAD; the bug is present in `0.12.4`.

py-spy on a wedged scheduler confirmed:
- the asyncio event loop is alive and iterating (`_run_once` timer advances every minute)
- `sched_count: 1` — exactly one timer, the main loop's own sleep
- the scheduler simply stops spawning new send tasks for affected `schedule_id`s

User-level mitigations — subclassing `TaskiqScheduler` and wrapping `on_ready` in `asyncio.wait_for` — don't help, because the wrapper runs *inside* the send task; if the send task itself never returns, the scheduler's `running_schedules` entry persists regardless. The only place a timeout can break the cycle is around the spawned send task itself, which is owned by the scheduler.

Filed as #626 with the full investigation.

## Change

1. New free function `send_with_timeout(scheduler, source, task, timeout)` in `taskiq/cli/scheduler/run.py` that wraps `send` in `asyncio.wait_for`. On `TimeoutError`, logs a WARNING and returns cleanly so the done_callback can clean up `running_schedules`. Non-timeout exceptions still propagate so existing observability surfaces (e.g., `SendTaskError` logs) are unchanged.
2. `SchedulerLoop.run` accepts `send_timeout: float | None = None`. When set, the spawned send task uses `send_with_timeout`; otherwise it uses `send` as before.
[...]

PR #669, opened 2026-09-01
fix: skip schedules with invalid cron_offset instead of crashing the scheduler

### LINKED ISSUE #668: A schedule with an invalid cron_offset string crashes the whole scheduler
### Taskiq version

0.12.6 (master)

### Python version

Python 3.12

### OS

Linux

### What happened?

`is_cron_task_now` in `taskiq/cli/scheduler/run.py` calls `ZoneInfo(offset)` outside the `try` block that converts `ValueError` into `CronValueError`, and `SchedulerLoop._is_schedule_ready_to_send` only catches `CronValueError`. A `cron_offset` that is not a valid IANA key raises `ZoneInfoNotFoundError` (a `KeyError` subclass) that escapes `SchedulerLoop.run` and kills the scheduler process. One malformed schedule stops all scheduling until the process is restarted.

```python
from datetime import datetime, timezone

from taskiq.cli.scheduler.run import SchedulerLoop
from taskiq.scheduler.scheduled_task import ScheduledTask

loop = SchedulerLoop.__new__(SchedulerLoop)
loop.cron_tasks_last_run = {}
loop.interval_tasks_last_run = {}
loop.time_tasks_last_run = {}

task = ScheduledTask(
    task_name="mod:cleanup",
    labels={},
    args=[],
    kwargs={},
    cron="* * * * *",
    cron_offset="UTC+3",  # natural spelling, but not an IANA key
)
loop._is_schedule_ready_to_send(task=task, now=datetime.now(tz=timezone.utc))
# -> zoneinfo.ZoneInfoNotFoundError escapes and kills SchedulerLoop.run
```

An invalid offset should get the same treatment as an invalid cron expression: log a warning and skip that schedule, keeping the loop alive.

The original timezone-offset request (#201) proposed per-schedule offsets as strings like `"+3"` and `"-1"`, so this spelling is a natural thing for users to write.

This is a different bug from #605 / #625. That change restores `timedelta` offsets from their serialized ISO-8601 form at model-validation time and does not touch `run.py`. A plain invalid timezone string still reaches `ZoneInfo()` at runtime. With the #625 patch applied, `cron_offset="UTC+3"` still crashes the loop, because `parse_cron_offset` leaves the string untouched (it is not a valid duration either).

### Relevant log output

```shell
zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key UTC+3'
```

```python
Not needed. The repro calls the scheduler loop directly.
```

juanmicl:
Closes: #668

`is_cron_task_now` called `ZoneInfo(offset)` outside the `try/except` that converts errors into `CronValueError`, and the scheduler loop only catches `CronValueError`. One schedule with a non-IANA `cron_offset` string (e.g. the natural spelling `"UTC+3"`) raised `ZoneInfoNotFoundError` out of `SchedulerLoop.run` and killed the whole scheduler process.

With this change a bad offset goes through the same path as an invalid cron expression: `CronValueError` is raised, the loop logs the existing `Cannot parse cron` warning and skips that schedule.

Two regression tests cover invalid offset strings: the natural spelling `"UTC+3"` and a typo'd timezone name (`"Europa/Madrid"`). Both fail on master with `ZoneInfoNotFoundError` and pass with this change.

Complements #625 (which fixes duration-string offsets at model-validation time but leaves the runtime guard unhandled).

Validation: full suite passes (`pytest -q`, 324 tests), `black`, `ruff` and `mypy` clean.

### DIFF
--- a/taskiq/cli/scheduler/run.py
+++ b/taskiq/cli/scheduler/run.py
@@ -99,7 +99,10 @@ def is_cron_task_now(
     # If timezone was specified as string we convert it timezone
     # offset and then apply.
     elif offset and isinstance(offset, str):
-        now = now.astimezone(ZoneInfo(offset))
+        try:
+            now = now.astimezone(ZoneInfo(offset))
+        except Exception as e:
+            raise CronValueError(e) from e
 
     try:
         return pycron.is_now(cron_value, now)
--- a/tests/cli/scheduler/test_is_cron_task_now.py
+++ b/tests/cli/scheduler/test_is_cron_task_now.py
@@ -75,3 +75,21 @@ def test_is_cron_task_now(
 def test_is_cron_task_now_invalid_cron() -> None:
     with pytest.raises(CronValueError):
         is_cron_task_now("invalid cron", datetime.now())
+
+
+def test_is_cron_task_now_invalid_offset_string() -> None:
+    with pytest.raises(CronValueError):
+        is_cron_task_now(
+            cron_value="* * * * *",
+            now=datetime.now(timezone.utc),
+            offset="UTC+3",
+        )
+
+
+def test_is_cron_task_now_unknown_timezone_name() -> None:
+    with pytest.raises(CronValueError):
+        is_cron_task_now(
+            cron_value="* * * * *",
+            now=datetime.now(timezone.utc),
+            offset="Europa/Madrid",
+        )


taskiq/cli/scheduler/run.py
```diff
@@ -99,7 +99,10 @@ def is_cron_task_now(
     # If timezone was specified as string we convert it timezone
     # offset and then apply.
     elif offset and isinstance(offset, str):
-        now = now.astimezone(ZoneInfo(offset))
+        try:
+            now = now.astimezone(ZoneInfo(offset))
+        except Exception as e:
```
danfimov, line 104: As far as I see `now.astimezone` can produce two types of exceptions here - `ZoneInfoNotFoundError` and `ModuleNotFoundError` (in case of importing invalid `tzdata.zoneinfo.Europa`). Can we use this types of error instead of broad Exception here?

tests/cli/scheduler/test_is_cron_task_now.py
```diff
@@ -75,3 +75,21 @@ def test_is_cron_task_now(
 def test_is_cron_task_now_invalid_cron() -> None:
     with pytest.raises(CronValueError):
         is_cron_task_now("invalid cron", datetime.now())
+
+
+def test_is_cron_task_now_invalid_offset_string() -> None:
```
danfimov, line 80: I think it can be one parametrized test

danfimov (changes requested):
Thank you for contribution)

Please fix couple minor things. And after that I think we can merge this MR.

taskiq/cli/scheduler/run.py
juanmicl: Done, switched to `except (ZoneInfoNotFoundError, ModuleNotFoundError)` with a short comment noting the missing-tzdata case.

danfimov:
Please run linters locally. I see that ruff is failing in CI

tests/cli/scheduler/test_is_cron_task_now.py
juanmicl: Merged into one parametrized test.

juanmicl:
I ran the linters locally before pushing, with both the locked ruff (0.16.5) and the latest one, and this branch passes. The errors in the CI lint job don't come from this PR's changes, they come from the merge with master.

#627 (b5d6033, merged after this PR was opened) introduced them: 5 auto-fixable `COM812` trailing comma issues in `tests/cli/scheduler/test_send_with_timeout.py`, and `C901` on `SchedulerLoop.run` (11 > 10) from the send_timeout branches. I verified by running ruff on the merge tree of this PR with current master: exactly the same 6 errors, none in the files this PR touches. They slipped in because the lint job in `test.yml` only runs on `pull_request` events, so pushes to master are never linted. Every PR currently open against master should show the same failure.

If it helps, I can send a small separate PR that fixes master's lint (apply the 5 auto-fixes and extract a small helper from `SchedulerLoop.run` to get back under the complexity limit), so this PR and the others go green. Or I can fold it into this PR if you prefer.

danfimov:
Yep, found the same problem. Will merge [this MR](https://github.com/taskiq-python/taskiq/pull/674) with fixes for pre-commit checks first, after that will rebase your MR and merge it too

danfimov (approved):

Issue #461, opened 2025-05-08
Scheduler SIGTERM/SIGINT Handler


lryan599:
I noticed that the scheduler does not have a handler set for sigterm and sigint, which leads to the scheduler as a pid 1 process not properly handling the sigterm signals sent by the kube to exit correctly when I deploy the scheduler in a container alone.

Is it possible to set the signal handler for taskiq scheduler just like worker?

robertaistleitner:
I just found the same issue because docker wasn't able to properly stop the container using the default SIGINT signal. It just waits for the timeout and hard kills it after the configured timeout.

PR #671, opened 2026-09-09
fix(scheduler): drain owned tasks before shutdown

Kuang-xianxin:
Cancelling `SchedulerLoop.run()` leaves its scheduled sends and background schedule refresh running. The CLI can then shut down the broker and sources while those operations still use them; the API can leave orphaned work in the application's event loop after cancellation has returned.

Cancel and await the loop's owned sends and refresh in `finally`, before returning control. Ignore a cancelled refresh in its completion callback so normal shutdown does not raise through the event-loop exception handler. This preserves existing CLI/API cancellation behavior and current master's optional `send_timeout` and startup rollback.

The event-controlled shutdown regression covers the loop, API and CLI, including timeout-wrapped sends for the loop and CLI. It verifies that the parent and shutdown hooks wait for asynchronous cleanup, checks the final cancellation state, and rejects callback errors. All five cases fail on unmodified current master (`8a1b04d`).

Validation after integrating current master: Windows / CPython 3.13.1, locked dependencies with all extras; 84 scheduler tests pass, all applicable changed-file hooks pass, and repository-wide Ruff/mypy pass (143 files for mypy). No live external broker was used, and the full repository test suite was not rerun for this update.

This complements #627's optional send timeout by handling operations owned by a loop that is exiting. It does not introduce a timeout or change the signal handling discussed in #461.


Kuang-xianxin:
Resolved the conflict with current master in ad69123, preserving the new optional send timeout and startup rollback alongside this PR's cancellation cleanup.

Extended the existing shutdown regression to cover timeout-wrapped sends through the loop and CLI. All five shutdown cases fail against unmodified current master (8a1b04d), and all 84 scheduler tests pass after the merge. Changed-file hooks, repository-wide Ruff and mypy (143 files) pass on Windows / Python 3.13.1. No live broker was used.

PR #674, opened 2026-09-16
chore: replace pre-commit with prek

danfimov:
- Replaced `pre-commit` with faster alternative `prek`
- Updated test libraries and pre-commit hooks


.pre-commit-config.yaml
```diff
@@ -29,6 +25,12 @@ repos:
             '--exclude-lines', 'amqp://guest:guest@localhost:5672',
         ]
 
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.16.7
+    hooks:
+      - id: ruff-check
+        args: [ "--fix", "taskiq", "tests", "docs/examples" ]
```
s3rius: I guess it shouldn't be `--fix`.

s3rius (changes requested):

.pre-commit-config.yaml
```diff
@@ -1,21 +1,17 @@
...
     hooks:
       - id: check-ast
+
+  - repo: builtin
+    hooks:
       - id: trailing-whitespace
       - id: check-toml
       - id: end-of-file-fixer
 
-  - repo: https://github.com/asottile/add-trailing-comma
```
danfimov, line 12: It's the same thing as ruff COM rules (they already enabled). So I removed it

taskiq/cli/scheduler/run.py
```diff
@@ -313,7 +313,7 @@ def _is_schedule_ready_to_send(
 
         return is_ready_to_send
 
-    async def run(
```
danfimov: I guess I missed this part in previously merged MRs somwhere. So now on master pre-commit hooks is red. I want to fix it.

tests/cli/scheduler/test_send_with_timeout.py
```diff
@@ -184,7 +192,9 @@ def __init__(self) -> None:
             self.calls: list[Any] = []
 
         async def on_ready(  # type: ignore[override]
```
s3rius: Why do you ignore? You can just add override decor.

s3rius (dismissed):

s3rius (approved):

PR #675, opened 2026-09-17
docs: add examples with ProgressTracker

### LINKED ISSUE #464: How to elegantly use result_backend.set_progress ？？
```python
@broker.task
async def heavy_task(key: str, val: str, context: Context = TaskiqDepends()) -> str:

    for i in range(10):
        # Simulate some work
        await asyncio.sleep(1)
        # Calculate progress
        progress = (i + 1) * 10
        # Send the progress
        # await context.set_progress(progress)

    await asyncio.sleep(10)
```

How to obtain the current task_id in the above code ？

Then set the progress using context.broker.result_backend.set_progress(). 

Because get_ progress returns None by default.

danfimov:
Closes #464 

P.S. ruff pre-commit is producing changes in master. So I added them here along with a new documentation to pass CI checks.


tests/cli/scheduler/test_send_with_timeout.py
```diff
@@ -31,7 +31,9 @@ def __init__(self, hang_seconds: float = 60.0) -> None:
         self.on_ready_calls = 0
 
     async def on_ready(  # type: ignore[override]
```
s3rius: I don't think we need those ignores. 

We can just add `override`. But it's just a small nitpick.

s3rius (dismissed):

tests/cli/scheduler/test_send_with_timeout.py
danfimov: Added override

s3rius (approved):

